PR #5, opened 2020-07-20
Add PuppetclassLookupKey model

ezr-ondrej:
Blocked by https://github.com/theforeman/foreman/pull/7808

PR #10, opened 2020-07-20
Add lookup value extensions

ezr-ondrej:
Blocked by #5


app/models/concerns/foreman_puppet_enc/puppet_lookup_value_extensions.rb
```diff
@@ -0,0 +1,13 @@
+module ForemanPuppetEnc
+  module PuppetLookupValueExtensions
+    extend ActiveSupport::Concern
+
+    included do
+      validate :value_present?, if: ->(v) { v.lookup_key.is_a?(ForemanPuppetEnc::PuppetclassLookupKey) }
```
ezr-ondrej, line 6: maybe some other distinction of type would be appropriate ... :thinking:
But that would need to be in core.

PR #12, opened 2020-07-27
Add config Groups Model

shiramax:


lib/foreman_puppet_enc/engine.rb
```diff
@@ -3,6 +3,8 @@ class Engine < ::Rails::Engine
     engine_name 'foreman_puppet_enc'
     isolate_namespace ForemanPuppetEnc
 
+    config.autoload_paths += Dir["#{config.root}/app/models/concerns"]
```
ezr-ondrej, line 6: Have you tested we need to do this?

app/models/foreman_puppet_enc/host/puppet_facet.rb
```diff
@@ -0,0 +1,25 @@
...
+      include ForemanPuppetEnc::HostCommon
+      include Facets::Base
+      include ::SelectiveClone
+
+      include_in_clone :config_groups, :host_config_groups
+
+      def clear_puppetinfo
+        unless environment
+          self.puppetclasses = []
+          self.config_groups = []
```
ezr-ondrej, line 15: is this working? :thinking:

ezr-ondrej: My bad, we've moved the associations to the `Facet`

db/migrate/20200803113903_migrate_hos_type_in_host_config_groups.foreman_puppet_enc.rb
```diff
@@ -0,0 +1,5 @@
+class MigrateHosTypeInHostConfigGroups < ActiveRecord::Migration[6.0]
+  def change
+    HostConfigGroups.where(:host_type => 'Host::Managed/Base').update_all(:host_type => 'ForemanPuppetEnc::HostFacet')
```
ezr-ondrej, line 3: I believe this won't be as easy, because we need to change the id column as well. Now we would assume, that the same `host_id` is referencing to the `facet_table`, but that is not the case.
Here is proposed (pseudo = untested) code:
```suggestion
  def up
    group_ids = HostConfigGroups.where(host_type: w%[Host::Base Host::Managed]).pluck(:config_group_id, :host_id)
    group_ids = group_ids.each_with_object(Hash.new([])) {|(cfg_id, host_id), memo| memo[host_id] += [cfg_id] }
    group_ids.each do |host_id, cfg_ids| do
      facet = ForemanPuppetEnc::HostFacet.create(host_id: host_id)
      HostConfigGroups.where(host_type: w%[Host::Base Host::Managed], host_id: host_id).update_all(host_type: 'ForemanPuppetEnc::HostFacet', host_id: facet.id)
    end
  end
```

shiramax:
@ezr-ondrej I think I've addressed all your comments,  let me know if there is something else :)

ezr-ondrej:
Thanks @shiramax !! :)

PR #14, opened 2020-07-27
Add Config Groups UI

shiramax:


lib/foreman_puppet_enc/engine.rb
```diff
@@ -3,6 +3,15 @@ class Engine < ::Rails::Engine
     engine_name 'foreman_puppet_enc'
     isolate_namespace ForemanPuppetEnc
 
+    config.autoload_paths += Dir["#{config.root}/app/controllers/concerns"]
+    config.autoload_paths += Dir["#{config.root}/app/helpers/concerns"]
+    config.autoload_paths += Dir["#{config.root}/app/models/concerns"]
+    config.autoload_paths += Dir["#{config.root}/app/overrides"]
```
ezr-ondrej, line 9: This should be loaded automatically I guess.

app/views/foreman_puppet_enc/config_groups/_config_group.html.erb
```diff
@@ -0,0 +1,34 @@
+<%= content_tag_for :li, config_group, (added ? :selected : nil), :id => "config_group_#{config_group.id}", :class => css_class.to_s do %>
+  <%= hidden_field_tag("#{type}[config_group_ids][]", config_group.id) if added == 'manually' %>
+  <span>
+   <%= link_to(icon_text('plus', config_group.name), '#', :rel=>'twipsy', :class => 'config_group_name',
+           :onclick => "tfm.classEditor.expandClassList($(this), '#cgroup_#{config_group.id}')",
```
ezr-ondrej, line 5: Where is this JS code living? Shouldn't we move it here as well?

app/views/foreman_puppet_enc/config_groups/_config_group.html.erb
```diff
@@ -0,0 +1,34 @@
...
+           :data => {:puppetclass_ids => config_group.puppetclass_ids,
+                    :'group-id' => config_group.id,
+                    :id => config_group.id,
+                    :type => type,
+                    :'original-title' => (obj.parent_config_groups.include?(config_group) ? _('included already from parent') : '' )
+                  }
+              ) %>
+  </span>
+  <%= link_to(" #{added ? _('Remove') : _('Add')}", '#', :rel=>'twipsy', :class => "btn btn-default btn-sm glyphicon glyphicon-#{added ? 'minus' : 'plus'}-sign",
+          :onclick => (added ? 'tfm.classEditor.removeConfigGroup(this)' : 'tfm.classEditor.addConfigGroup(this)'),
```
ezr-ondrej, line 15: The same (JS)

config/routes.rb
```diff
@@ -10,5 +14,5 @@
 end
 
 Foreman::Application.routes.draw do
-  mount ForemanPuppetEnc::Engine, at: '/foreman_puppet'
+  mount ForemanPuppetEnc::Engine, at: '/foreman_puppet_enc'
```
ezr-ondrej, line 17: this was intentional, but I don't mind eighter way :)

lib/foreman_puppet_enc/engine.rb
```diff
@@ -13,17 +22,55 @@ class Engine < ::Rails::Engine
...
+          permission :create_config_groups, { :config_groups => %i[new create],
+                                              :"api/v2/config_groups" => [:create] }
+          permission :edit_config_groups, { :config_groups => %i[edit update],
+                                            :"api/v2/config_groups" => [:update] }
+          permission :destroy_config_groups, { :config_groups => [:destroy],
+                                               :"api/v2/config_groups" => [:destroy] }
+        end
 
-        # add menu entry
+        role 'ForemanPuppetEnc', [:view_foreman_puppet_enc]
```
ezr-ondrej, line 41: Do we want extra role for PuppetEnc handling?

lib/foreman_puppet_enc/engine.rb
```diff
@@ -13,17 +22,55 @@ class Engine < ::Rails::Engine
...
+        divider :top_menu, :parent => :configure_menu, :after => :configure, :caption => N_('Puppet ENC')
+
+        # add menu entry
+        add_menu_item :top_menu, :puppet_enc, {
+          :caption => 'Config Groups',
+          :engine => ForemanPuppetEnc::Engine, :parent => :configure_menu, :after => :ansible,
+          :url_hash => { :controller => 'foreman_puppet_enc/config_groups', :action => :index }
+        }
         # add dashboard widget
+        widget 'foreman_puppet_enc_widget', name: N_('Foreman plugin template widget'), sizex: 4, sizey: 1
```
ezr-ondrej, line 53: we don't have any widget for now, do we?

ezr-ondrej (commented):
Could we rebase and get in the JS ? :)

ezr-ondrej:
Can you rebase please? :)

lib/foreman_puppet_enc/engine.rb
shiramax: nope, I will delete this line.

lib/foreman_puppet_enc/engine.rb
shiramax: I think we should, what do you think?

ezr-ondrej: I would keep the roles as they are now in core and there is no special role for managing puppet :thinking:
If to have such role, I would say it should be a role `ConfigurationManager` or something like that, but otherwise I'd keep things simple.

lib/foreman_puppet_enc/engine.rb
```diff
@@ -5,6 +5,10 @@ class Engine < ::Rails::Engine
 
     config.paths['db/migrate'] << 'db/migrate_foreman' if Gem::Dependency.new('', '>= 2.3').match?('', SETTINGS[:version])
 
+    if Gem::Dependency.new('', '>= 2.2').match?('', SETTINGS[:version])
+      config.paths['db/migrate'] << 'db/migrate_foreman'
+    end
```
ezr-ondrej, line 10: we have this already line ^^ above

ezr-ondrej:
I'm missing something like: https://github.com/theforeman/foreman_statistics/blob/master/webpack/index.js#L10 to set the `tfm.classEditor` (I guess it is ok to override the core implementation?)

lib/foreman_puppet_enc/engine.rb
```diff
@@ -15,19 +15,53 @@ class Engine < ::Rails::Engine
...
         requires_foreman '>= 2.0.0'
-
         # Add Global JS file for extending foreman-core components and routes
         register_global_js_file 'fills'
 
         # Add permissions
-        # security_block :foreman_puppet_enc do
-        # end
+        security_block :foreman_puppet_enc do
+          permission :view_config_groups, { :config_groups => %i[index auto_complete_search welcome],
```
ezr-ondrej, line 23:
```suggestion
          permission :view_config_groups, { :'foreman_puppet_enc/config_groups' => %i[index auto_complete_search welcome],
```

lib/foreman_puppet_enc/engine.rb
```diff
@@ -15,19 +15,53 @@ class Engine < ::Rails::Engine
...
+                                            :"api/v2/config_groups" => %i[index show],
```
ezr-ondrej, line 24: this can be removed for now, but if you want to bring it before we have the actual API controllers, lets prefix it as well.
```suggestion
                                            :"foreman_puppet_enc/api/v2/config_groups" => %i[index show],
```

lib/foreman_puppet_enc/engine.rb
```diff
@@ -15,19 +15,53 @@ class Engine < ::Rails::Engine
...
+                                            :'foreman_puppet_enc/react' => [:index] }
+          permission :create_config_groups, { :config_groups => %i[new create],
```
ezr-ondrej, line 26:
```suggestion
          permission :create_config_groups, { :'foreman_puppet_enc/config_groups' => %i[new create],
```
And so on... ^_^

ezr-ondrej (changes requested):
Forgotten permission prefix and the JS registration, but apart of that I believe we are set :+1:

ezr-ondrej:
We can rebase and enable the tests in `test/controllers/config_groups_controller_tests.rb`, that should now stop failing :+1:

app/views/foreman_puppet_enc/hosts/puppet.erb
```diff
@@ -0,0 +1 @@
+Welcome to <b>ForemanPuppetEnc</b>
```
ezr-ondrej, line 1: Hmm :thinking: sound like warming welcome, but not sure where we are using it :smiley_cat:

shiramax: @ezr-ondrej I removed this file :) thanks

ezr-ondrej (approved):
Yup this works, thanks @shiramax :+1:

ezr-ondrej:
Please create the Extraction PR :upside_down_face:

PR #30, opened 2020-09-28
Add smart class parameters API

ezr-ondrej:
Adds SmartclassParametersController to handle API requests.
Tests has been moved to use factories instead of fixtures.


ezr-ondrej:
:green_apple: :tada:

PR #88, opened 2020-12-21
Fixes #31538 - pull in fix for distinct association

ezr-ondrej:
From core PR https://github.com/theforeman/foreman/pull/8215.


ezr-ondrej:
We should add tests for that :thinking:

PR #119, opened 2021-02-19
Taxable Environment

ezr-ondrej:
Add environemnt_ids and environments to taxonomies and TaxHost.

### DIFF
--- a/app/models/concerns/foreman_puppet/extensions/taxonomy.rb
+++ b/app/models/concerns/foreman_puppet/extensions/taxonomy.rb
@@ -14,6 +14,14 @@ module ForemanPuppet
         new.environments = environments
         new
       end
+
+      def environment_ids do
+        if ignore?('ForemanPuppet::Environment')
+          return ForemanPuppet::Environment.pluck(:id)
+        else
+          super()
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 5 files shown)


timogoebel:
Is this still relevant?

ezr-ondrej:
> Is this still relevant?

Depends, do we want to get Environment into the taxonomy tabs?

app/models/concerns/foreman_puppet/extensions/taxonomy.rb
```diff
@@ -14,6 +14,14 @@ def dup
         new.environments = environments
         new
       end
+
+      def environment_ids do
```
timogoebel, line 18:
```suggestion
      def environment_ids
```

?

timogoebel:
> Depends, do we want to get Environment into the taxonomy tabs?

Well, I guess we want to keep the behavior before extraction. If I checked correctly, this functionality is still left in core. But I checked in a rush.

My question is: Do we need this in the 1.0 release?

ezr-ondrej:
> My question is: Do we need this in the 1.0 release?

My understanding is no. It is regression from before extracted, but OTH no plugin entities are added to the Taxonomy edit :shrug:

stejskalleos:
Closing the PR since there haven't been any updates for almost two years.

PR #120, opened 2021-02-22
More of missing host tests

ezr-ondrej:


timogoebel:
Is this still relevant?

ezr-ondrej:
More tests never hurts

PR #121, opened 2021-03-10
Popup test fixes

ezr-ondrej:
In https://github.com/theforeman/foreman/pull/8364 we've changed the
popup initialization and it afected our test suite.


tbrisker (approved):
Thanks @ezr-ondrej !

tbrisker (commented):
It looks like tests are failing now?

ezr-ondrej:
> It looks like tests are failing now?

It's using an old develop so the popup is already initialized and the title is not there anymore, let me fix that and update the test branch :)

PR #122, opened 2021-03-10
Add puppet related setting

ezr-ondrej:


nadjaheitmann:
@ezr-ondrej What about this PR? I saw that there is some development in Foreman managing settings at the moment. Shall I look into this PR or better wait until things are settled in Foreman?

ezr-ondrej:
I'm not sure about this one, I'm trying to resolve the setting issue for few months already, but you can se we're hardly half way through: https://projects.theforeman.org/issues/29939.

The main problem is puppet category doesn't have all the puppet settings and it has some that are used elsewhere, so this needs to be merge prior the extraction: https://github.com/theforeman/foreman/pull/8383, which is hard without the setting refactoring.

So we've got currently two solutions:
- keep all the settings in core and not extract them with puppet (very weird, but working solution, this is currently implemented in puppet_extraction branch) and wait for 3.1 where the setting refactoring will be done.
- do the reordering the old way, there might be dragons that we do not forsee, because moving setting from category to non-existent category might :bomb: 
- push the setting refactoring harder

nadjaheitmann:
Sounds like we should wait with the settings until reorganization is done. Otherwise, we will need to do more work than necessary to get the settings up and running.

ezr-ondrej:
Merging as it is just moving the settings here.

PR #128, opened 2021-04-01
Add puppetmaster to compatibility properties

ezr-ondrej:
In https://github.com/theforeman/foreman/pull/8423 we've decided to not
safeguard by plugin, but by availability of the `puppetmaster` method.

This adds this method to backward compatible properties available on `Host::Managed`.


timogoebel:
This is waiting on the core PR to get in. If it gets accepted, this looks great.

ezr-ondrej:
This should possibly introduce both `puppetserver` and `puppetmaster` and deprecating the `puppetmaster` as per https://github.com/theforeman/foreman/pull/8423#pullrequestreview-626117860

ezr-ondrej:
Should be done within #146

PR #110, opened 2021-01-31
Add Puppetclass GraphQL

ezr-ondrej:

PR #134, opened 2021-04-20
Add GraphQL

kamils-iRonin:
@ezr-ondrej I rebased https://github.com/theforeman/foreman_puppet/pull/110 and continue working on GraphQL

TODO:
- create host mutation


lib/foreman_puppet/engine.rb
```diff
@@ -1,5 +1,24 @@
...
+            return if args.first == :puppetclasses && args.second.node_type.to_s == 'Types::Puppetclass'
+
+            super
+          end
+        end
+
+        GraphQL::Types::Relay::BaseObject.extend(BaseObjectClassMethodPath)
+      end
+    end
+
```
kamils-iRonin, line 21: a hacky way to skip the core types when the plugin is installed and to avoid duplicates.... any better ideas?

kamils-iRonin: I agree, the priority should be to make sure the plugin works with the `puppet_extraction` branch

app/models/foreman_puppet/puppet_facet_common.rb
```diff
@@ -1,8 +1,9 @@
 module ForemanPuppet
   # Common methods between host and hostgroup
   # mostly for template rendering consistency
-  module HostCommon
+  module PuppetFacetCommon
```
ezr-ondrej, line 4: While I totaly agree with this change, it should be in separate PR, right?

lib/foreman_puppet/engine.rb
ezr-ondrej: I agree, let's not bother by making it work with 2.5 as GraphQL is not stable API, so it's reasonable to say it works only with 3.0, if you need it, just wait with installing the plugin :)

ezr-ondrej (changes requested):
I can see you're implementing the proxies, while that's awesome, it should have separate PR IMHO :)

kamils-iRonin:
> I can see you're implementing the proxies, while that's awesome, it should have separate PR IMHO :)

@ezr-ondrej I extracted the smart proxy code from this PR and think it's ready

app/models/foreman_puppet/puppet_facet_common.rb
kamils-iRonin: opened https://github.com/theforeman/foreman_puppet/pull/158

ezr-ondrej (approved):
I'm good here. @timogoebel any comments?

timogoebel:
@kamils-iRonin: Sorry, this needs a rebase, now.

timogoebel:
Tests fail. Looks unrelated, but we should check it out before merging imho.

nadjaheitmann:
Tests fail because of this commit in foreman `develop` branch:
https://github.com/theforeman/foreman/commit/16226afc7a804d5a416f6fdc77a9e7b85970a600

It changes an error message which we check in the unit tests. I experoenced the same in my PR.

timogoebel:
> It changes an error message which we check in the unit tests. I experoenced the same in my PR.

What do we need to do to make the tests pass again?

kamils-iRonin:
> What do we need to do to make the tests pass again?

@timogoebel It's fixed (https://github.com/theforeman/foreman_puppet/compare/7752e312def5673629ef87e110c696e27d41dd57..de81a2a11b06714b4fc7df94a9ea16652885a9f1) I'm working on tests against `puppet_extraction` branch now

nadjaheitmann:
> It's fixed

Do you mind opening a separate PR for this fix? I think, it is not related to GraphQL and would be nice to have in `master` such that the pipeline stops failing.

kamils-iRonin:
> Do you mind opening a separate PR for this fix? I think, it is not related to GraphQL and would be nice to have in `master` such that the pipeline stops failing.

@nadjaheitmann sure, opened https://github.com/theforeman/foreman_puppet/pull/161

timogoebel:
And we need another rebase. Thanks for fixing this!

.github/workflows/ruby_tests.yml
```diff
@@ -83,7 +83,7 @@ jobs:
         node-version: [12]
         include:
           - foreman-core-branch: puppet_extraction
```
ezr-ondrej, line 85: Do our branches differs?
I'm trying to keep it up to date by constatnt rebases and I'll include any changes from reviews also in that branch.
But hopefully we soon can change this to develop, so I dont mind much :)

kamils-iRonin: I temporarily changed to my branch where I extracted the graphql stuff and rebased with develop. Before this PR is merged we should update your branch, but let's wait to see if the tests pass

kamils-iRonin: opened https://github.com/ezr-ondrej/foreman/pull/66

ezr-ondrej (approved):
Great work @kamils-iRonin ! :+1:  Thanks for taking care of all the issues on the way and sorry for causing most of them...

PR #139, opened 2021-05-12
Test against develop

### LINKED ISSUE #91: Test against develop
We are testing against custom branch now, we should switch back to testing against develop as most of the fixes are in the develop now.

kamils-iRonin:
closes #91

### DIFF
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -66,7 +66,7 @@ jobs:
         run: bundle exec rake spec
   test_ruby:
     runs-on: ubuntu-latest
-    needs: rubocop
+    # needs: rubocop
     env:
       BUNDLE_WITHOUT: journald:development:console:libvirt
     services:
@@ -77,7 +77,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        foreman-core-branch: [hostgroup_facet-tempate_input]
+        foreman-core-branch: [2.5-stable, develop]
         ruby-version: [2.5, 2.6]
         node-version: [12]
     steps:
@@ -86,7 +86,7 @@ jobs:
         uses: nanasess/setup-chromedriver@master
       - uses: actions/checkout@v2
         with:
-          repository: ezr-ondrej/foreman
+          repository: theforeman/foreman
           ref: ${{ matrix.foreman-core-branch }}
       - uses: actions/checkout@v2
         with:
--- a/app/controllers/foreman_puppet/api/v2/hostgroup_classes_controller.rb
+++ b/app/controllers/foreman_puppet/api/v2/hostgroup_classes_controller.rb
@@ -28,6 +28,10 @@ module ForemanPuppet
           process_response @hostgroup_class.destroy_all
         end
 
+        def resource_class
+          ForemanPuppet::HostgroupClass
+        end
+
         private
 
         def find_hostgroup
--- a/app/models/concerns/foreman_puppet/extensions/host.rb
+++ b/app/models/concerns/foreman_puppet/extensions/host.rb
@@ -6,8 +6,13 @@ module ForemanPuppet
       included do
         prepend PrependedMethods
 
-        env_assoc = reflect_on_association(:environment)
-        env_assoc&.instance_variable_set(:@class_name, 'ForemanPuppet::Environment')
+        unless ForemanPuppet.extracted_from_core?
+          env_assoc = reflect_on_association(:environment)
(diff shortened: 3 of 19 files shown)


lib/foreman_puppet.rb
```diff
@@ -1,5 +1,5 @@
 module ForemanPuppet
-  FOREMAN_EXTRACTION_VERSION = '2.6'.freeze
+  FOREMAN_EXTRACTION_VERSION = '2.7'.freeze
```
ezr-ondrej, line 2: Why? This will be correct as once we extract puppet 2.6 => 3.0 and 3.0 >= 2.7, but it just looks weird.
We can go with 3.0 directly :)

app/models/concerns/foreman_puppet/extensions/host.rb
```diff
@@ -9,6 +9,9 @@ module Host
         env_assoc = reflect_on_association(:environment)
         env_assoc&.instance_variable_set(:@class_name, 'ForemanPuppet::Environment')
 
+        host_classes_assoc = reflect_on_association(:host_classes)
+        host_classes_assoc&.instance_variable_set(:@class_name, 'ForemanPuppet::HostClass')
```
ezr-ondrej, line 13: both should be `unless extracted`

app/controllers/foreman_puppet/api/v2/hostgroup_classes_controller.rb
```diff
@@ -28,6 +28,10 @@ def destroy
           process_response @hostgroup_class.destroy_all
         end
 
+        def resource_class
+          ForemanPuppet::HostgroupClass
```
ezr-ondrej, line 32: Why are these needed? :thinking:

ezr-ondrej (commented):
Few nits, but apart of that great job!
Do we want to merge before release for 2.5?

app/controllers/foreman_puppet/api/v2/hostgroup_classes_controller.rb
kamils-iRonin: to return `ForemanPuppet::HostgroupClass` instead of `::HostgroupClass` from core

```ruby
From: /Users/kamilszubrycht/Documents/Development/foreman_puppet/app/controllers/foreman_puppet/api/v2/hostgroup_classes_controller.rb:33 ForemanPuppet::Api::V2::HostgroupClassesController#resource_class:

    31: def resource_class
    32:   binding.pry
 => 33:   ForemanPuppet::HostgroupClass
    34: end

[1] pry(#<ForemanPuppet::Api::V2::HostgroupClassesController>)> super
=> HostgroupClass(id: integer, hostgroup_id: integer, puppetclass_id: integer, hostgroup_puppet_facet_id: integer)
[2] pry(#<ForemanPuppet::Api::V2::HostgroupClassesController>)> ForemanPuppet::HostgroupClass
=> ForemanPuppet::HostgroupClass(id: integer, hostgroup_id: integer, puppetclass_id: integer, hostgroup_puppet_facet_id: integer)
```

lib/foreman_puppet.rb
kamils-iRonin: bumped into `3.0`, on current develop `SETTINGS[:version].notag` returns `2.6.0`

kamils-iRonin:
> Do we want to merge before release for 2.5?

I think it makes sense to merge it before release for 2.5

.github/workflows/ruby_tests.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         run: bundle exec rake spec
   test_ruby:
     runs-on: ubuntu-latest
-    needs: rubocop
+    # needs: rubocop
```
ezr-ondrej, line 69: Yeah, I've got no idea what's wrong with rubocop :(

app/controllers/foreman_puppet/api/v2/hostgroup_classes_controller.rb
ezr-ondrej: oh right, in API we have wrong namespace, so it doesn't get guessed correctly.

.github/workflows/ruby_tests.yml
kamils-iRonin: I have no idea either, I'm investigating

ezr-ondrej: I'll take a look into that in follow-up

ezr-ondrej (approved):
Thanks @kamils-iRonin ! :+1:

Issue #132, opened 2021-04-06
Host puppetclass assignment results in error and broken class list


laugmanuel:
While editing a host and assigning puppet related configuration (puppetserver, environment, classes), the save fails with:
```
11:23:39 rails.1   | 2021-04-06T11:23:39 [I|app|0e511529] Started PATCH "/hosts/41" for ::1 at 2021-04-06 11:23:39 +0200
11:23:39 rails.1   | 2021-04-06T11:23:39 [I|app|0e511529] Processing by HostsController#update as */*
11:23:39 rails.1   | 2021-04-06T11:23:39 [I|app|0e511529]   Parameters: {"utf8"=>"✓", "authenticity_token"=>"s6wHzWYI1dXVcg+acn1PQ1lHE4QTjBS/a+UI4nB2jqYZyBKV6OdWTQTDX9vHCHVG7Ud/F8BFAjq8s0LZuYFcHQ==", "host"=>{"name"=>"louis-tatem", "hostgroup_id"=>"", "puppet_attributes"=>{"environment_id"=>"1"}, "puppet_proxy_id"=>"1", "puppetclass_ids"=>["", "1"], "managed"=>"true", "progress_report_id"=>"[FILTERED]", "interfaces_attributes"=>{"0"=>{"_destroy"=>"0", "mac"=>"aa:bb:cc:dd:ee:ff", "identifier"=>"", "name"=>"louis-tatem", "domain_id"=>"2", "subnet_id"=>"1", "ip"=>"", "ip6"=>"", "managed"=>"1", "primary"=>"1", "provision"=>"1", "tag"=>"", "attached_to"=>"", "id"=>"41"}}, "architecture_id"=>"1", "operatingsystem_id"=>"1", "medium_id"=>"1", "ptable_id"=>"97", "pxe_loader"=>"", "disk"=>"", "root_pass"=>"[FILTERED]", "is_owned_by"=>"4-Users", "enabled"=>"1", "model_id"=>"", "expired_on"=>"", "comment"=>"", "overwrite"=>"false"}, "id"=>"41"}
11:23:39 rails.1   | 2021-04-06T11:23:39 [D|tax|0e511529] Current location set to Default Location
11:23:39 rails.1   | 2021-04-06T11:23:39 [D|tax|0e511529] Current organization set to Default Organization
11:23:39 rails.1   | 2021-04-06T11:23:39 [D|tax|0e511529] Current organization set to none
[...]

PR #140, opened 2021-05-14
Patch param modules earlier

ezr-ondrej:
We can end up with controllers getting required prior to_prepare and
that breaks patching the host_params.

Refs #132

### DIFF
--- a/app/models/concerns/foreman_puppet/extensions/host.rb
+++ b/app/models/concerns/foreman_puppet/extensions/host.rb
@@ -61,6 +61,12 @@ module ForemanPuppet
       end
 
       module PrependedMethods
+        # TODO: we can drop this once extracted_from_core?
+        def validate_association_taxonomy(association_name)
+          return if association_name.to_sym == :environment
+          super
+        end
+
         def provisioning_template(opts = {})
           opts[:environment_id] ||= puppet&.environment_id
           super(opts)
--- a/app/controllers/concerns/foreman_puppet/extensions/parameters_host.rb
+++ b/app/controllers/concerns/foreman_puppet/extensions/parameters_host.rb
@@ -34,16 +34,9 @@ module ForemanPuppet
         end
 
         def process_deprecated_environment_params!(params)
-          env_id = env_name = env = nil
-          if ForemanPuppet.extracted_from_core?
-            env_id = params.delete(:environment_id)
-            env_name = params.delete(:environment_name)
-            env = params.delete(:environment)
-          else
-            env_id = params[:environment_id]
-            env_name = params[:environment_name]
-            env = params[:environment]
-          end
+          env_id = params.delete(:environment_id)
+          env_name = params.delete(:environment_name)
+          env = params.delete(:environment)
 
           return unless env_id || env_name || env
           ::Foreman::Deprecation.api_deprecation_warning('param host[environment_*] has been deprecated in favor of host[puppet_attributes][environment_*]')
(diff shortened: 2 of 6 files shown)


ezr-ondrej:
@laugmanuel could you check with your setup it fixes #132?

laugmanuel:
@ezr-ondrej it does not seem to help. Now I do see the Puppetclasses in GUI (still with the same layout problem), but I can't see them using `rake console`.
Removing a Puppetclass using GUI is also still not possible.

`rake console`:
```ruby
[46] pry(main)> Host.find_by(name: 'gail-galluzzo.foo.example.local').puppetclasses
=> []
````

API:
```sh
❯ curl -u admin:admin http://localhost:5000/api/hosts/gail-galluzzo.foo.example.local | jq '.puppetclasses'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  3451    0  3451    0     0   7042      0 --:--:-- --:--:-- --:--:--  7028
[
  {
    "id": 2,
    "name": "test",
    "module_name": "test"
  },
  {
    "id": 1,
    "name": "test::config",
    "module_name": "test"
  }
]

❯ curl -u admin:admin http://localhost:5000/api/hosts/gail-galluzzo.foo.example.local/puppetclasses | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   151    0   151    0     0    480      0 --:--:-- --:--:-- --:--:--   480
{
  "total": 0,
  "subtotal": 0,
  "page": 1,
  "per_page": 20,
  "search": null,
  "sort": {
    "by": null,
    "order": null
  },
  "results": {}
}
```

ezr-ondrej:
Yeah didn't realize, we are breaking the current methods, that will be dropped and API redirected to new enpoints, but your finds should work for now.

ezr-ondrej:
What about now, I didn't want pollute the Host model by puppet related methods much, but I'm fine with it for 2.5 :)
And it should fix your finds.

laugmanuel (approved):
This works like a charm now :) Together with #143 this seems to resolve all issues found in #132

app/models/concerns/foreman_puppet/extensions/host.rb
```diff
@@ -61,6 +61,12 @@ def search_by_config_group(_key, operator, value)
       end
 
       module PrependedMethods
+        # TODO: we can drop this once extracted_from_core?
+        def validate_association_taxonomy(association_name)
+          return if association_name.to_sym == :environment
+          super
+        end
```
kamils-iRonin, line 68:
```suggestion
        # TODO: we can drop this once extracted_from_core?
        unless ForemanPuppet.extracted_from_core?
          def validate_association_taxonomy(association_name)
            return if association_name.to_sym == :environment
            super
          end
        end
```

ezr-ondrej: we can do that, but I'd rather drop this method once we extract puppet from core.
If we accidently don't do that, this should be `noop`, so I feel the check is not necessary.

kamils-iRonin: if we add the `puppet_extracted` core branch to CI, won't there be a problem with tests without this check?

ezr-ondrej: Nope, this method will be noop

ezr-ondrej: there would be no such association on the `Host::Managed`

PR #143, opened 2021-05-14
Fixes ability to remove class

ezr-ondrej:


laugmanuel (approved):
@ezr-ondrej, I've tested this and it works as expected. As noted in https://github.com/theforeman/foreman_puppet/pull/140#issuecomment-841194274 the UI is still a bit broken (layout wise as seen in https://github.com/theforeman/foreman_puppet/issues/132), but the buttons do work. The ENC also responds to the changes and adds/removes the classes accordingly.

PR #145, opened 2021-05-17
Switch rubocop to 2.7

ezr-ondrej:
It seems rubocop was ignoring the ArgumentAlignment rule up until now.

### DIFF
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 2.5
+          ruby-version: 2.7
           bundler-cache: true
       - name: Run rubocop
         run: bundle exec rubocop -P --format github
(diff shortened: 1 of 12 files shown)


.github/workflows/ruby_tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 2.5
+          ruby-version: 2.7
```
kamils-iRonin, line 22: wdyt about using ruby `2.7` also for tests? (and drop `2.5` and `2.6`)

kamils-iRonin (approved):

timogoebel (approved):

ezr-ondrej:
CP 0.1: #149

.github/workflows/ruby_tests.yml
ekohl: Ruby 2.5 is still used on Debian 10 and Ubuntu 18.04 and both are supported platforms.

PR #146, opened 2021-05-17
Add smart proxies

ezr-ondrej:


ezr-ondrej:
Or should we add the proxies on Host?

timogoebel:
> Or should we add the proxies on Host?

I'm not sure if I'm familiar enough with the topic, but can we delegate the methods from `Host` model to the facet object? Just for compatibility.

ezr-ondrej:
> I'm not sure if I'm familiar enough with the code, but can we delegate the methods from `Host` model to the facet object? Just for compatibility.

We definitelly can, but I'm not sure if that's the best way forward as there is specialized API for plugins to add their proxies to `Host` model. So I'm not sure what's the prefered way. I think that today it's not really being used, but the proxies would be great to have together I think, so something like "All host proxies" queries would be easy to do.

timogoebel:
> as there is specialized API for plugins to add their proxies to `Host` model

Yep, I'm quite familiar with the API as I initially wrote it. :-D

I'm all for using that API to register the proxies, that makes a lot of sense imho.

timogoebel:
Looks like this is still part of core. Do we need this PR?

ezr-ondrej:
I believe this should get in, but only for Foreman 3.1 release.

timogoebel:
> I believe this should get in, but only for Foreman 3.1 release.

@ezr-ondrej: Do you mind creating a milestone for this so we have better visibility? Thanks.

stejskalleos:
Closing the PR since there haven't been any updates for almost two years.

PR #151, opened 2021-05-21
Fix apidoc position of Puppet plugin

### LINKED ISSUE #147: API inconsistencies
There some inconsistencies in the API regarding

- documentation/apidoc missing
- parameters (some not listed, some optional although they should not be)
- documentation showing up in a separate block in apidoc

nadjaheitmann:
This fixes #147

### DIFF
--- a/app/controllers/foreman_puppet/api/v2/puppet_base_controller.rb
+++ b/app/controllers/foreman_puppet/api/v2/puppet_base_controller.rb
@@ -1,10 +1,9 @@
 module ForemanPuppet
   module Api
     module V2
-      class BaseController < ::Api::V2::BaseController
+      class PuppetBaseController < ::Api::V2::BaseController
         resource_description do
-          api_version '2'
-          api_base_url '/foreman_puppet/api'
+          api_version 'v2'
         end
 
         before_action :show_deprecation_for_core_routes
--- a/app/controllers/foreman_puppet/api/v2/config_groups_controller.rb
+++ b/app/controllers/foreman_puppet/api/v2/config_groups_controller.rb
@@ -1,8 +1,12 @@
 module ForemanPuppet
   module Api
     module V2
-      class ConfigGroupsController < BaseController
+      class ConfigGroupsController < ::ForemanPuppet::Api::V2::PuppetBaseController
         include ForemanPuppet::Parameters::ConfigGroup
+
+        resource_description do
+          api_base_url '/foreman_puppet/api'
+        end
 
         wrap_parameters ConfigGroup, include: config_group_params_filter.accessible_attributes(parameter_filter_context)
 
--- a/app/controllers/foreman_puppet/api/v2/environments_controller.rb
+++ b/app/controllers/foreman_puppet/api/v2/environments_controller.rb
@@ -1,12 +1,11 @@
 module ForemanPuppet
   module Api
     module V2
-      class EnvironmentsController < BaseController
+      class EnvironmentsController < ::ForemanPuppet::Api::V2::PuppetBaseController
         include ForemanPuppet::Api::ImportPuppetclassesCommonController
         include ForemanPuppet::Parameters::Environment
 
         resource_description do
-          api_version '2'
           api_base_url '/foreman_puppet/api'
         end
 
--- a/app/controllers/foreman_puppet/api/v2/host_classes_controller.rb
+++ b/app/controllers/foreman_puppet/api/v2/host_classes_controller.rb
@@ -1,7 +1,11 @@
 module ForemanPuppet
   module Api
     module V2
-      class HostClassesController < V2::BaseController
+      class HostClassesController < ::ForemanPuppet::Api::V2::PuppetBaseController
+        resource_description do
+          api_base_url '/foreman_puppet/api'
+        end
+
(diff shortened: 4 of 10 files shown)


app/controllers/foreman_puppet/api/v2/base_controller.rb
```diff
@@ -3,7 +3,8 @@ module Api
     module V2
       class BaseController < ::Api::V2::BaseController
         resource_description do
-          api_version '2'
+          resource_id 'foreman_puppet_base'
```
ezr-ondrej, line 6: Where is the id of the resource used?

nadjaheitmann: It is used by Apipie to distinguish between the classes and get the documentation correctly, as far as I can tell.

nadjaheitmann: I should change the resource_id of the `BaseController` and the one of the `LookupsCommonController` to have a `pp_` prefix such that all of the resource_ids are consistent.

nadjaheitmann:
With this PR, the API entries appear in addition to the core entries. Is this okay or should they replace the core entries?

timogoebel:
> With this PR, the API entries appear in addition to the core entries. Is this okay or should they replace the core entries?

I believe when the plugin is installed, the core models don't work anymore and the endpoints from this plugin should be used.

nadjaheitmann:
> I believe when the plugin is installed, the core models and controllers don't work anymore and the endpoints from this plugin should be used.
As of now, @ezr-ondrej said that the old routes are redirected to the new ones. But in any case, I think it makes more sense to remove the old apidoc entries or at least mark them as deprecated as it gives the impression that both can be used in parallel. I will fix that.

timogoebel:
I'm wondering if we can add something to the CI to verify all relevant fields are properly documented.

nadjaheitmann:
> I'm wondering if we can add something to the CI to verify all relevant fields are properly documented.

This would be very helpful indeed as comparing all API entries manually is very tedious. I will think about it. Let me know if you have an idea how to do it.

nadjaheitmann:
> As of now, @ezr-ondrej said that the old routes are redirected to the new ones.

Correction: Old routes are not redirected, yet, but end up at the core API. Do we want to fix this? 
```
[I|app|e93963d1] Started GET "/api/puppetclasses" for 192.168.121.1 at 2021-0    6-01 06:45:45 +0000$
[I|app|e93963d1] Processing by Api::V2::PuppetclassesController#index as JSON
```

With this PR, the documentation of the core API is overwritten, such that only new routes are shown. Moreover, `resource_id` stays the same as in core for all commands, such that e.g. Hammer can automatically use the new routes instead of the old ones.

The PR fixes #147 that were caused mostly by `resource_id` inconsistencies between Foreman core and the plugin.

timogoebel:
> Correction: Old routes are not redirected, yet, but end up at the core API. Do we want to fix this?

Yes, I think we should. We could also just disable the core routes. I think we should just make sure this is actually worth the effort. I can see a lot of tests in core break if we implement this.

> Let me know if you have an idea how to do it.

I think it should be fairly easy to get all relevant API controllers (e.g. via the routes)  and all the actions. We then need possible parameters that we could get from the parameter filters. And to finish it up, the documentation dsl should give us a list of all already documented parameters.

nadjaheitmann:
> Yes, I think we should. We could also just disable the core routes. I think we should just make sure this is actually worth the effort. I can see a lot of tests in core break if we implement this.

I will open a new issue for this. 
 
> I think it should be fairly easy to get all relevant API controllers (e.g. via the routes) and all the actions. We then need possible parameters that we could get from the parameter filters. And to finish it up, the documentation dsl should give us a list of all already documented parameters.

Let me give it a try. I guess it makes sense to have the tests as part of this PR? But I have not seen tests for apidoc completion before, so could be a new issue as well...

timogoebel:
> But I have not seen tests for apidoc completion before, so could be a new issue as well...

I'd recommend a separate issue.

nadjaheitmann:
Is there anything that needs fixing for this PR?

timogoebel (approved):
Looks good to me.

app/controllers/foreman_puppet/api/v2/puppet_base_controller.rb
```diff
@@ -1,10 +1,9 @@
 module ForemanPuppet
   module Api
     module V2
-      class BaseController < ::Api::V2::BaseController
+      class PuppetBaseController < ::Api::V2::BaseController
```
kamils-iRonin, line 4: Is there any other reason for renaming the controllers, or do you just feel it's a better name?

nadjaheitmann: The reason is that the `resource_id` of any controller is generated from the name of the controller. As there already exist a controller named `BaseController` in Foreman core, there is a name clash. It is okay for the controllers that should be overwritten anyways, but not for the ones that should remain in Foreman core.

kamils-iRonin (approved):

kamils-iRonin:
Thanks @nadjaheitmann 👍

PR #154, opened 2021-06-08
Re-route API core routes to plugin controllers

### LINKED ISSUE #152: Redirect core routes to use plugin routes
Refers to https://github.com/theforeman/foreman_puppet/pull/151#issuecomment-851874564 .

As of now, old API routes are still active. 
```
[I|app|e93963d1] Started GET "/api/puppetclasses" for 192.168.121.1 at 2021-06-01 06:45:45 +0000$
[I|app|e93963d1] Processing by Api::V2::PuppetclassesController#index as JSON
```
Two possibilities:
1) redirect the routes for now (seems easier in terms of implementation and more user friendly for current API users) 
2) disable routes (we be done eventually, even if we implement the redirection for now)

nadjaheitmann:
Fixes #152

### DIFF
--- a/config/initializers/api_reroute.rb
+++ b/config/initializers/api_reroute.rb
@@ -1,0 +1,13 @@
+Foreman::Application.routes.prepend do
+  namespace :api, defaults: { format: 'json' } do
+    scope '(:apiv)', module: :v2, defaults: { apiv: 'v2' }, apiv: /v2/, constraints: ApiConstraints.new(version: 2, default: true) do
+      resources :puppetclasses, controller: '/foreman_puppet/api/v2/puppetclasses', except: %i[new edit]
+      resources :environments, controller: '/foreman_puppet/api/v2/environments', except: %i[new edit]
+      resources :hosts, controller: '/foreman_puppet/api/v2/hosts', only: []
+      resources :hostgroups, controller: '/foreman_puppet/api/v2/hostgroups', only: %i[]
+      resources :config_groups, controller: '/foreman_puppet/api/v2/config_groups', except: %i[new edit]
+      resources :smart_class_parameters, controller: '/foreman_puppet/api/v2/smart_class_parameters', except: %i[new edit]
+      resources :override_values, controller: '/foreman_puppet/api/v2/override_values', except: %i[new edit]
+    end
+  end
+end


nadjaheitmann:
Seems there is something wrong with rubocop?

sbernhard:
> Seems there is something wrong with rubocop?

I guess, another PR is need to fix the rubocop complains.

nadjaheitmann:
> I guess, another PR is need to fix the rubocop complains.
Yeah, I will look into this.

nadjaheitmann:
Right now, I have simply re-directed the controllers, which seems to work. I have seen that (re)routing is often restricted for specific actions. What is the best practice for this case?

ezr-ondrej:
We have this in core in the extraction PR, where we reroute if the plugin is installed. Though this might be a better way to go as in core we are checking wheter a plugin is installed in routes file, depends how long we want to keep the rerouting. Do we wanna get rid of it in two versions? then lets keep it in core, if we wanna do this permanently, lets merge this.

timogoebel:
> if we wanna do this permanently, lets merge this.

I vote for having it as part of the plugin. This keeps all puppet related code here and we don't have to do core changes in the future.

nadjaheitmann:
> Do we wanna get rid of it in two versions? then lets keep it in core, if we wanna do this permanently, lets merge this.

We might be a more flexible how we want to handle the routing if we merge it here and Foreman core routes are cleaner. Then we can avoid something like this:
https://github.com/theforeman/foreman/blob/develop/config/routes/api/v2.rb#L207

nadjaheitmann:
If we were to merge this, is there anything else that needs to be done for the PR?

timogoebel:
I haven't done a thorough review, but basically, this looks good.
I believe we need to check if permission wise anything needs to be done.
In foreman_dlm for example we have this check as part of the CI. I suggest we also add it here. The check makes sure that for all routes correct access permissions are defined.

Test that we should add to the CI:

https://github.com/dm-drogeriemarkt/foreman_dlm/blob/f230261ce5be8d98d199333a5015854a87b7995d/.github/workflows/ci.yml#L69

API to define permissions for routes (as an example):
https://github.com/dm-drogeriemarkt/foreman_dlm/blob/f230261ce5be8d98d199333a5015854a87b7995d/lib/foreman_dlm/engine.rb#L32-L35

timogoebel:
Looks good. Can we drop the WIP-Label?

timogoebel:
Thanks!

PR #155, opened 2021-06-09
Update rubocop to version 1.16.0

nadjaheitmann:


timogoebel:
We should check if we want to pin the rubocop version so our CI doesn't break when a new version is released.

nadjaheitmann:
I was actually thinking about that. Might be a good idea to update rubocop eventually, although I would think that this is something forgotten easily.  

But let's go for it... There is 1.16.1 already and tests break again:
https://github.com/theforeman/foreman_puppet/runs/2782876698?check_suite_focus=true

PR #156, opened 2021-06-09
Pin rubocop version to 1.16.0

nadjaheitmann:
Prevent rubocop updates to break the CI (randomly).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,4 +20,5 @@ end
 gemspec
 
 gem 'rspec-rails', '~> 4.0.1', groups: %i[development test]
+gem 'rubocop', '<= 1.16.0', groups: %i[development rubocop]
 gem 'theforeman-rubocop', '~> 0.1.0.pre', groups: %i[development rubocop]


ezr-ondrej:
Could we pin this in https://github.com/theforeman/theforeman-rubocop/blob/fc0ef342248f189750a1295e78628b56fbe4eb25/theforeman-rubocop.gemspec#L23 ? We are hitting this in all plugins using theforeman-rubocop

nadjaheitmann:
Makes sense. Do you want to bump a new `theforeman-rubocop` version, then?

ezr-ondrej:
> Makes sense. Do you want to bump a new `theforeman-rubocop` version, then?

Yes, I'll cut new release, which shoul be included here automatically.

nadjaheitmann:
> Yes, I'll cut new release, which shoul be included here automatically.

Okay. But `~> 1.15.0` won't catch version `1.16.0`, will it.

nadjaheitmann:
> Okay. But `~> 1.15.0` won't catch version `1.16.0`, will it.

Would it make sense to define an explicit upper bound for rubocop, e.g. `1.16.0` and have a periodic update of this upper bound? For something like `~> 1.15.0`, you never know whether there might be a `1.15.1` or not.

ezr-ondrej:
> you never know whether there might be a `1.15.1` or not.

Well in theory it should not be breaking change, and thus I would expect it not break the rubocop rules, if 1.16.1 did, it was a bad relase on rubocop part, but I would not expect it to be the case too often and `~> x.y.z` is supposed to be "I want to get bugfixes" what is what we want I believe.

nadjaheitmann:
> `~> x.y.z` is supposed to be "I want to get bugfixes" what is what we want I believe.
IMHO, this makes sense for stable plugins. For this plugin, it would be nice to have a stable rubocop version. Right now, there is a new rubocop release approx. every 2 weeks and it broke twice in a row. I would prefer focusing on new features and bug fixes  rather than being distracted by broken CI ;) So let's make a decision and decide what to do with the PR.

ezr-ondrej (approved):
Make sense, we can play with rubocop once extraction is done, lets pin for now :+1:
Thanks @nadjaheitmann

PR #157, opened 2021-06-21
Add access permissions check for >= 3.0

nadjaheitmann:
Add access permission test to CI and fix permission for the test to succeed, base for #154

### DIFF
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -131,6 +131,10 @@ jobs:
           bundle exec rake test:foreman_puppet
         env:
           PUPPET_EXTRACTED: ${{ matrix.foreman-core-branch == 'puppet_extraction' && '1' }}
+      - name: Run access permission test
+        if: ${{ ( startsWith(matrix.foreman-core-branch, '3') || matrix.foreman-core-branch == 'puppet_extraction' ) && github.event_name != 'push' }}
+        run: |
+          bundle exec rake test TEST="test/unit/foreman/access_permissions_test.rb"
       - name: Upload test.log
         uses: actions/upload-artifact@v2
         if: failure() && github.event_name != 'push'


lib/foreman_puppet/register.rb
```diff
@@ -59,89 +59,89 @@
...
+                                                      lookup_values: [:destroy],
+                                                      'foreman_puppet/api/v2/smart_class_parameters': [:destroy],
+                                                      'foreman_puppet/api/v2/override_values': %i[create update destroy] },
       resource_type: 'ForemanPuppet::PuppetclassLookupKey'
   end
 
   security_block :puppet_environments do
-    permission :view_environments, { 'foreman_puppet/environments': %i[index show auto_complete_search welcome],
-                                     'foreman_puppet/api/v2/environments': %i[index show] },
+    permission :puppet_view_environments, { 'foreman_puppet/environments': %i[index show auto_complete_search welcome],
```
timogoebel, line 101: Do we need db migrations to change the permission names?

timogoebel: On second thought, do we want to change the name here? I'm wondering if the `puppet` prefix breaks things or not.

nadjaheitmann: Permissions test for master:
https://github.com/ATIX-AG/foreman_puppet/runs/2882491673?check_suite_focus=true

To me it looks like there is a name clash between original permissions and the plugin and the original permissions are overwritten: https://github.com/theforeman/foreman/blob/3f7804fd96d380010b340f2a954f9cb42a60c551/app/registries/foreman/plugin.rb#L330 
That is why I tried renaming. Correct me if I am wrong.

nadjaheitmann: But maybe overwriting the existing permission is exactly what we want and we should let the tests fail for Foreman < 3.0 ?

timogoebel: Yep, sounds good. With the plugin we do want to override core functionality, so I'd be ok with that.

nadjaheitmann: I have changed the PR to contain the permissions check starting from a 3.X stable branch.

.github/workflows/ruby_tests.yml
```diff
@@ -123,6 +123,10 @@ jobs:
         if: github.event_name != 'push'
         run: |
           bundle exec rake test:foreman_puppet
+      - name: Run access permission test
+        if: ${{ startsWith(matrix.foreman-core-branch, '3') && github.event_name != 'push' }}
```
timogoebel, line 127: Is this code currently reachable?

nadjaheitmann: The if condition is being checked, but the test does not run as we do not have a foreman core branch starting with '3' (, yet).

.github/workflows/ruby_tests.yml
```diff
@@ -123,6 +123,10 @@ jobs:
...
+        run: |
+          bundle exec rake test TEST="test/unit/foreman/access_permissions_test.rb"
```
ezr-ondrej, line 129: btw we don't need to test all the routes, we can just test the engine routes like:
https://github.com/theforeman/foreman_statistics/blob/master/test/unit/foreman_statistics/access_permissions_test.rb
OR
https://github.com/theforeman/foreman_bootdisk/blob/master/test/unit/access_permissions_test.rb

nadjaheitmann: There already is https://github.com/theforeman/foreman_puppet/blob/master/test/unit/foreman_puppet/access_permissions_test.rb

nadjaheitmann: Shall we close the PR if the test is already there?

ezr-ondrej: this effort shown that the test doesn't test the integration with core routes.

So however we decide to rewrite core API routes (eighter directly in core, as I have that in my PR or with your PR here) that would be untested. So this test still brings value in testing that override.

nadjaheitmann: Right. Shall we move ahead and merge it, then?

PR #161, opened 2021-06-28
Fix SmartClassParameters API test

kamils-iRonin:
there is a difference in the error message depending on the Foreman version (introduced https://github.com/theforeman/foreman/commit/16226afc7a804d5a416f6fdc77a9e7b85970a600)


nadjaheitmann:
Thanks @kamils-iRonin

PR #162, opened 2021-06-30
Use default :matchers_inheritance setting

kamils-iRonin:
the tests fail with an `ActiveRecord::RecordInvalid: Validation failed: Name has already been taken` error, I guess it's after merging of https://github.com/theforeman/foreman/pull/8437


nadjaheitmann:
Thanks @kamils-iRonin

timogoebel:
Thanks!

PR #164, opened 2021-07-06
Fix config group links

nadjaheitmann:


ezr-ondrej (approved):
Thanks @nadjaheitmann.
Care to add some specs to verify this? I know it's not very straight forward, so I dont insist :)

nadjaheitmann:
@ezr-ondrej If you point me to an example how to verify links using specs, I can give it a try :)

ezr-ondrej:
You can take a look at `spec/views/foreman_puppet_enc/environments/index.html.erb_spec.rb`, and use `have_selector` to verify the link. but the specs have quite artificial test environments, so it's not 100% testing that the links will work in prod if they work in the spec.

ezr-ondrej:
Thanks @nadjaheitmann

PR #166, opened 2021-07-19
GraphQL fixes

kamils-iRonin:


timogoebel (approved):

timogoebel:
Thanks!

ezr-ondrej:
Do we want to cherry-pick this? I do not agree with `2.6` version btw, it is nonexistent version :)

kamils-iRonin:
> Do we want to cherry-pick this?

Yes, I think it should be cherry-picked. 

> I do not agree with `2.6` version btw, it is nonexistent version :)

Changed to `2.6` to make tests on `develop` pass (`SETTINGS[:version].notag` returned `2.6.0`). I think we can set `3.0` back now.

ezr-ondrej:
> Changed to `2.6` to make tests on `develop` pass (`SETTINGS[:version].notag` returned `2.6.0`). I think we can set `3.0` back now.

Yeah, I get it, good thing it's in separate commit :)

ezr-ondrej:
CP 1.0: 48d6ee7

PR #170, opened 2021-07-22
Fix smart proxy import

nadjaheitmann:
POST /foreman_puppet/api/smart_proxies/:id/import_puppetclasses
POST /foreman_puppet/api/smart_proxies/:smart_proxy_id/environments/:id/import_puppetclasses 
POST /foreman_puppet/api/environments/:environment_id/smart_proxies/:id/import_puppetclasses

were not working although they appear in the apidoc


timogoebel:
Thanks!

PR #172, opened 2021-07-26
Update translations 

nadjaheitmann:
https://community.theforeman.org/t/plugin-maintainers-action-required-request-to-extract-strings-for-translations/24601

Or shall we wait with updating until 1.0.0 is released?


timogoebel:
Thanks!



> Or shall we wait with updating until 1.0.0 is released?

Can't hurt, can it?

PR #173, opened 2021-07-27
Fix smart proxy importclasses with invalid environment

nadjaheitmann:
If you pass an invalid `environment_id` via API, you get `{"message":"No changes to your environments detected"}`.

### DIFF
--- a/app/controllers/concerns/foreman_puppet/api/import_puppetclasses_common_controller.rb
+++ b/app/controllers/concerns/foreman_puppet/api/import_puppetclasses_common_controller.rb
@@ -74,7 +74,13 @@ module ForemanPuppet
         end
 
         unless @environments.any?
-          render_message(_('No changes to your environments detected'))
+          # check if environment id/name was valid at all
+          if @env_id && \
+             Environment.authorized(:view_environments).where(name: @env_id).or(Environment.authorized(:view_environments).where(id: @env_id)).empty?
+            render_message(_('The requested environment cannot be found.'))
+          else
+            render_message(_('No changes to your environments detected'))
+          end
           return false
         end
 
--- a/test/controllers/foreman_puppet/api/v2/environments_controller_test.rb
+++ b/test/controllers/foreman_puppet/api/v2/environments_controller_test.rb
@@ -251,7 +251,8 @@ module ForemanPuppet
               ProxyAPI::Puppet.any_instance.stubs(:environments).returns(%w[env1 env2])
               classes_env1 = { 'a' => Foreman::ImporterPuppetclass.new('name' => 'a') }
               classes_env2 = { 'b' => Foreman::ImporterPuppetclass.new('name' => 'b') }
-              ProxyAPI::Puppet.any_instance.stubs(:classes).returns(classes_env1.merge(classes_env2))
+
+              ProxyAPI::Puppet.any_instance.stubs(:classes).with('invalid').returns({})
               ProxyAPI::Puppet.any_instance.stubs(:classes).with('env1').returns(classes_env1)
               ProxyAPI::Puppet.any_instance.stubs(:classes).with('env2').returns(classes_env2)
             end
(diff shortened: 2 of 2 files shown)


app/controllers/concerns/foreman_puppet/api/import_puppetclasses_common_controller.rb
```diff
@@ -74,7 +74,13 @@ def changed_environments
         end
 
         unless @environments.any?
-          render_message(_('No changes to your environments detected'))
+          # check if environment id/name was valid at all
+          if @env_id && (Environment.authorized(:view_environments).find_by(id: @env_id) || \
+             Environment.authorized(:view_environments).find_by(name: @env_id))
+            render_message(_('No changes to your environments detected'))
+          else
+            render_message(_('The environment with id %s cannot be found.') % @env_id)
```
timogoebel, line 82: Is it possible to inject html here?
I might be overly cautious here.
@kamils-iRonin: WDYT?

test/controllers/foreman_puppet/api/v2/environments_controller_test.rb
```diff
@@ -270,6 +270,18 @@ class EnvironmentsControllerTest < ActionController::TestCase
               assert_response :success
             end
 
+            test 'should render error message when invalid environment is set' do
+              env = 'invalid'
+              # assert_difference(-> { ForemanPuppet::Puppetclass.unscoped.count }, 0) do
```
timogoebel, line 275: The debug code should be removed prior merge.

test/controllers/foreman_puppet/api/v2/environments_controller_test.rb
```diff
@@ -279,7 +291,7 @@ class EnvironmentsControllerTest < ActionController::TestCase
               assert_response :success
             end
 
-            context 'ignored entvironments or classes are set' do
+            context 'ignored environments or classes are set' do
```
timogoebel, line 294: Nice. :)

test/controllers/foreman_puppet/api/v2/environments_controller_test.rb
nadjaheitmann: Definitely. Right now, the test is failing, though. Any idea why?

timogoebel: I don't have time to look here myself. @kamils-iRonin: Do you mind chiming in here?

app/controllers/concerns/foreman_puppet/api/import_puppetclasses_common_controller.rb
```diff
@@ -74,7 +74,13 @@ def changed_environments
         end
 
         unless @environments.any?
-          render_message(_('No changes to your environments detected'))
+          # check if environment id/name was valid at all
+          if @env_id && (Environment.authorized(:view_environments).find_by(id: @env_id) || \
```
ezr-ondrej, line 78: this is wrong, what about when we do not pass the `env_id`?
Shouldn't we just detect whether the `@env_id` is set or not and tailor the message according to that?

app/controllers/concerns/foreman_puppet/api/import_puppetclasses_common_controller.rb
ezr-ondrej, line 82: The fact, that it is not in DB, should not be a failure, we can be importing a new environment.
If it doesn't exists, you might be just trying to import new env, that you've misspelled.

Wouldn't it be easier just detect whether `env_id` is set and if it is let the message be something like:
`No changes for environment <env_id> detected`?

nadjaheitmann: > The fact, that it is not in DB, should not be a failure, we can be importing a new environment.

This is checked earlier. If a new environment is detected before, it will skip that code.  

> what about when we do not pass the env_id

You are right, I need to adjust the message if there is no `@env_id`.

> Wouldn't it be easier just detect whether env_id is set and if it is let the message be something like:

Easier, I agree. Just not sure if the user realizes that he/she misspelled the environment. IMHO it might be clearer to explicitly tell that it is not existing.

test/controllers/foreman_puppet/api/v2/environments_controller_test.rb
kamils-iRonin: the test fails because the puppet proxy api returns mocked classes for any environment name, see:

https://github.com/theforeman/foreman_puppet/blob/3069d9022c41b69252932a5679d1735e28adb84f/test/controllers/foreman_puppet/api/v2/environments_controller_test.rb#L254

the test should pass after replacing this line with:

```ruby
ProxyAPI::Puppet.any_instance.stubs(:classes).with('invalid').returns({})
```

(tested with `develop` core branch)

nadjaheitmann: Thanks a lot, much appreciated @kamils-iRonin

app/controllers/concerns/foreman_puppet/api/import_puppetclasses_common_controller.rb
kamils-iRonin: I'm not sure, we should be safe here 🤔

nadjaheitmann: Not sure, either. The easiest fix is to make the message more general such as `The requested environment cannot be found.`

nadjaheitmann:
How shall we proceed here?

timogoebel:
> How shall we proceed here?

Looks good to me. Anything to add?

kamils-iRonin (approved):

timogoebel:
Thanks!

PR #176, opened 2021-08-03
Fixes #33203 - use global constant, not module

ezr-ondrej:


timogoebel:
Do we want a test for this?

ares:
Probably, at least I can confirm it works and fixes the issue :-)

PR #180, opened 2021-08-05
Fixes #33217 - Add :all_puppetclasses to hostgroup

xprazak2:


ezr-ondrej:
I'm not sure we want to do this. It never was an public API and we are making it an public API by this (it's added to the safemode), I'd rather seen it removed from Host as well, I'm not sure why I've added it (probably as temporary solution).
Could we add it to the Extensions::Hostgroup instead? Ideally for both Host and Hostgroup.

xprazak2:
Updated, methods have been moved to extensions for both Host and Hostgroup.

app/models/concerns/foreman_puppet/extensions/host.rb
```diff
@@ -3,6 +3,8 @@ module Extensions
     module Host
       extend ActiveSupport::Concern
 
+      include ForemanPuppet::Extensions::HostCommon
```
ezr-ondrej, line 6:
```suggestion
      include ForemanPuppet::Extensions::HostCommon if ForemanPuppet.extracted_from_core?
```

app/models/concerns/foreman_puppet/extensions/hostgroup.rb
```diff
@@ -3,6 +3,8 @@ module Extensions
     module Hostgroup
       extend ActiveSupport::Concern
 
+      include ForemanPuppet::Extensions::HostCommon
```
ezr-ondrej, line 6: also only if extracted

ezr-ondrej (approved):
Seems to work :+1:
Thanks @xprazak2

ezr-ondrej:
CP 1.0: 25a0789

PR #183, opened 2021-09-10
Drop extracted_from_core? and require Foreman 3.0

ekohl:
I must admit I only looked for extracted_from_core and am not very familiar with the source code.

The reasoning is that the version check fails with Foreman 3.0.0 GA. See https://community.theforeman.org/t/foreman-3-0-0-ga-release-process/25151/2 for more details.


lib/foreman_puppet/register.rb
```diff
@@ -212,11 +183,9 @@
...
 
-      if ForemanPuppet.extracted_from_core?
-        context.add_pagelet :main_tab_fields,
-          partial: 'hosts/foreman_puppet/form_main_tab_fields',
-          resource_type: resource_type,
-          priority: 100
+      context.add_pagelet :main_tab_fields,
+        partial: 'hosts/foreman_puppet/form_main_tab_fields',
+        resource_type: resource_type,
       end
```
ezr-ondrej, line 189: This is redundant `end`

app/controllers/foreman_puppet/puppetclasses_controller.rb
```diff
@@ -14,7 +14,6 @@ class PuppetclassesController < ApplicationController
 
     helper ForemanPuppet::PuppetclassLookupKeysHelper
 
-    # TODO: extracted_from_core?
```
ezr-ondrej, line 17: I'd like to keep this todo as `filter for facets needs improving`

ezr-ondrej (commented):
Just two nits and rubocop will need fixing. We will also need to test only against 3.0 foreman (bumping GH action workflow `ruby_tests.yml` from `2.5-stable` to `3.0-stable`

PR #184, opened 2021-09-10
Add core migration one version after the extraction

ezr-ondrej:


lib/foreman_puppet/engine.rb
```diff
@@ -33,7 +33,7 @@ def argument(name, type, *rest, loads: nil, **kwargs, &block)
     engine_name 'foreman_puppet'
     isolate_namespace ForemanPuppet
 
-    config.paths['db/migrate'] << 'db/migrate_foreman' if Gem::Dependency.new('', ">= #{ForemanPuppet::FOREMAN_EXTRACTION_VERSION}").match?('', SETTINGS[:version])
+    config.paths['db/migrate'] << 'db/migrate_foreman' if Gem::Dependency.new('', "> #{ForemanPuppet::FOREMAN_EXTRACTION_VERSION}").match?('', SETTINGS[:version])
```
evgeni, line 36: this means that it will be included again when we do 3.0.1, which is not what your intention is, right? you want it to be included on 3.1+, as that version removes the migrations from core?

ekohl (commented):
Why do we need to keep this logic? It's not like we need to care about 2.5 anymore. In #183 I suggested to drop it.

ezr-ondrej:
> Why do we need to keep this logic? It's not like we need to care about 2.5 anymore. In #183 I suggested to drop it.

I've drop the use of the constant for the version. Though the idea is, we will remove the core migrations from Foreman and remove them from DB `migration_schema`, so in case you do not install plugin, it will look like they've never existed. If you do install the plugin, it will take over the ownership of the migrations.

lib/foreman_puppet.rb
```diff
@@ -1,5 +1,6 @@
 module ForemanPuppet
   FOREMAN_EXTRACTION_VERSION = '2.6'.freeze
```
evgeni, line 2: I think in the 1.0 branch, you have a commit changing this to 3.0, probably should be aligned in master too ;)

ezr-ondrej: it should, but given we want to drop it ASAP from master, I think it's ok not to :)

lib/foreman_puppet/engine.rb
```diff
@@ -33,7 +33,7 @@ def argument(name, type, *rest, loads: nil, **kwargs, &block)
     engine_name 'foreman_puppet'
     isolate_namespace ForemanPuppet
 
-    config.paths['db/migrate'] << 'db/migrate_foreman' if Gem::Dependency.new('', ">= #{ForemanPuppet::FOREMAN_EXTRACTION_VERSION}").match?('', SETTINGS[:version])
+    config.paths['db/migrate'] << 'db/migrate_foreman' if Gem::Dependency.new('', ">= #{ForemanPuppet::FOREMAN_DROP_MIGRATIONS_VERSION}").match?('', SETTINGS[:version])
```
ehelms, line 36: In theory, since you are saying `>=`, you are wanting this inclusion of migrations to happen with 3.1 which is effectively current nightly. However, we are not seeing this execute on nightly and we would face the problem where this does not execute until the actual release given:

```
irb(main):001:0> Gem::Dependency.new('', ">= 3.0").match?('', '3.0.0.rc2')
=> false
irb(main):002:0> Gem::Dependency.new('', ">= 3.0").match?('', '3.1.0-develop')
=> false
irb(main):003:0> Gem::Dependency.new('', ">= 3.0").match?('', '3.0.0')
=> true
```

So yea, this would be a stop gap fix for 3.0, but it would still result in incorrect behavior IMO. I am OK with fixing something to get 3.0 happily out the door, but it seems like maybe you should just remove this line all together for the 3.0 release?

ezr-ondrej: we can get this in and remove this line in `foreman_puppet 1.0`

ezr-ondrej: I've added notag and opened a https://github.com/theforeman/foreman_puppet/pull/186 for version 1.0

PR #192, opened 2021-09-17
Force the insert on migration

### LINKED ISSUE #191: Migration fails with PGSql adapter error on some environments
From: https://community.theforeman.org/t/tfm-rubygem-foreman-puppet-noarch-db-migrate-error/25322

`ActiveRecord::ConnectionAdapters::PostgreSQLAdapter does not support skipping duplicates`
Which points to badly configrued Postgresql, but we cant do anything about it, so we should try fixing it for such environments.

Probably relevant: https://github.com/rails-sqlserver/activerecord-sqlserver-adapter/issues/847

ezr-ondrej:
If we do not force inserts, Rails tries to skip duplicates, which is not supported on some environments.
This makes the migration flaky and fail on badly configured environmets for no reason.

Fixes #191


ezr-ondrej:
CP 1.0: dd96ad2

PR #193, opened 2021-09-17
Added missing Puppet CA trait

lzap:
Previously, there was a trait named `with_puppet_orchestration` that had both puppet and puppetca proxies associated. Now there is no such trait, this adds it.

Would it be possible to backport this into plugin version for Foreman 3.0 so I could use it in many of discovery tests for 3.0 branch?


ezr-ondrej:
CP 1.0: 3b24a38

PR #194, opened 2021-09-17
Use subject in pagelet template directly

ezr-ondrej:


lzap:
Can confirm this fixes discovery alongside with https://github.com/theforeman/foreman_puppet/pull/193 thanks. I would also need this in 3.0 version of the plugin.

ezr-ondrej:
CP 1.0: f11a77a

PR #195, opened 2021-09-22
Adds Puppetclass migrations

ezr-ondrej:
Migrations of puppetclass taken from core and added here.


ezr-ondrej:
We will need to test this after the migrations are removed from core, merging now and will fix when core does remove them.

PR #196, opened 2021-09-23
Fixes #171 - Resolve smart proxy api inconsistencies

### LINKED ISSUE #171: Smart proxy apidoc inconsistency
Environments and smart proxies both have the same API endpoint for importing puppet classes. I tried to prepend the `smart_proxy` path with `foreman_puppet' for #170 but so far no success. It is a bit tricky because it is defined as concern. Any ideas?

image
image

nadjaheitmann:
Needs the [concern ](https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/api/import_puppetclasses_common_controller.rb) to be removed from core to work properly. 

PR in core: https://github.com/theforeman/foreman/pull/8791(removes slightly more then needed for this to work)


app/controllers/foreman_puppet/api/v2/environments_controller.rb
```diff
@@ -2,7 +2,7 @@ module ForemanPuppet
   module Api
     module V2
       class EnvironmentsController < ::ForemanPuppet::Api::V2::PuppetBaseController
-        include ForemanPuppet::Api::ImportPuppetclassesCommonController
+        include ForemanPuppet::Api::ImportPuppetclassesEnvController
```
ezr-ondrej, line 5: Why are we including it here? I thought we don't want it on Environments controller?

app/controllers/concerns/foreman_puppet/extensions/api_smart_proxies_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      api :POST, '/environments/:environment_id/smart_proxies/:id/import_puppetclasses', N_('Import puppet classes from puppet proxy for an environment')
+      param :id, :identifier, required: true
+      param :smart_proxy_id, String, required: false
+      param :environment_id, String, required: false
+      param :dryrun, :bool, required: false
+      param :except, String, required: false, desc: N_('Optional comma-delimited string'\
+                                                       "containing either 'new', 'updated', or 'obsolete'"\
+                                                       'that is used to limit the imported Puppet classes')
+
+      include ForemanPuppet::Api::ImportPuppetclassesEnvController
```
ezr-ondrej, line 18: If I'm correct and we don't want to have it in the environments controller anymore, can we have the method here instead of a concern? :)

.github/workflows/ruby_tests.yml
```diff
@@ -77,12 +77,17 @@ jobs:
...
       matrix:
         foreman-repo: [theforeman/foreman]
         foreman-core-branch: [2.5-stable, develop]
         ruby-version: [2.7]
         node-version: [12]
+        include:
+          - foreman-core-branch: remove_puppet_api_routes
+            foreman-repo: 'ATIX-AG/foreman'
+            ruby-version: 2.7
+            node-version: 12
```
ezr-ondrej, line 90: can we leave this out? now? :)

app/controllers/concerns/foreman_puppet/extensions/api_smart_proxies_controller.rb
nadjaheitmann: Yeah, let me fix this... I had a routing error which I now resolved and the code became much clearer.

ezr-ondrej (approved):
Make sense to me :+1:

ezr-ondrej:
Thanks @nadjaheitmann ! :+1:

PR #199, opened 2021-09-24
Hosts count for environments

ezr-ondrej:


lib/foreman_puppet/engine.rb
```diff
@@ -80,6 +80,8 @@ def argument(name, type, *rest, loads: nil, **kwargs, &block)
       ::TemplateCombination.include ForemanPuppet::Extensions::TemplateCombination
       ::ProvisioningTemplate.include ForemanPuppet::Extensions::ProvisioningTemplate
 
+      ::HostCounter.prepend ForemanPuppet::Extensions::HostCounter
```
timogoebel, line 83: Should we do `include` here

and have a module Extensions that is prepended in the concern?

ezr-ondrej: I don't see the value in that... if the only method we are patching needs to be prepended :thinking:

timogoebel:
Oh, and can we have a small test for this?

lib/foreman_puppet/engine.rb
timogoebel: It would just be consistency. But I'm fine if we leave it like this for now.

timogoebel (approved):

timogoebel:
Thanks!

ezr-ondrej:
CP 1.0: c00ae5b

PR #201, opened 2021-09-27
Move all core routes to plugin

nadjaheitmann:
Prerequisite for https://github.com/theforeman/foreman/pull/8791


ezr-ondrej (approved):
Thanks @nadjaheitmann

PR #203, opened 2021-10-06
Fixes #33645 - remove react-intl from package.json

Ron-Lavi:
`react-intl` is now part of foreman-vendor 8.15.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,25 +21,22 @@
     "url": "http://projects.theforeman.org/projects/foreman_puppet/issues"
   },
   "peerDependencies": {
-    "@theforeman/vendor": ">= 6.0.0"
+    "@theforeman/vendor": "^8.15.0"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -21,25 +21,22 @@
...
+    "@theforeman/vendor-dev": "^8.15.0",
     "babel-eslint": "^10.0.3",
     "eslint": "^6.7.2",
+    "jed": "^1.1.1",
     "prettier": "^1.19.1",
     "stylelint-config-standard": "^18.0.0",
     "stylelint": "^9.3.0"
-  },
-  "dependencies": {
-    "jed": "^1.1.1",
```
laviro, line 42: jed is being used here only as a devDep, so we can remove now the whole dependencies section

laviro:
@ezr-ondrej can you take a look? :)

PR #208, opened 2021-10-14
Support Foreman 3.0

ezr-ondrej:


ezr-ondrej:
Probably not crucial, 3.0 worked, we do not need the new version to work with 3.0

nadjaheitmann:
You are cherry-picking the most important changes to version 1 anyways, aren't you? We need to update readme and dependencies, if there is a breaking change, though.

ezr-ondrej:
Will not happen anyways :shrug:

PR #209, opened 2021-10-19
Always return an array

### LINKED ISSUE #206: foreman_puppet/app/models/foreman_puppet/host_puppet_facet.rb: parent_config_groups() should not return 'nil'
source: foreman_puppet/app/models/foreman_puppet/host_puppet_facet.rb, line51.
method parent_config_groups() should not return a nil value, since that results in an exception in callers. This happens in the puppet node classifier code path whenever a hostgroup is encountered that does not have any puppet plugin data.

The patch indicated below fixes this issue.

```
    def parent_config_groups
          return [] unless host.hostgroup
          host.hostgroup.puppet&.all_config_groups || []
                                                   ^^^^^^
    end
```

ezr-ondrej:
Ensures the return value of parent_config_groups is always array.

Fixes #206

Thanks @pkranenburg for reporting and sending the patch :)

### DIFF
--- a/app/models/foreman_puppet/host_puppet_facet.rb
+++ b/app/models/foreman_puppet/host_puppet_facet.rb
@@ -48,7 +48,7 @@ module ForemanPuppet
 
     def parent_config_groups
       return [] unless host.hostgroup
-      host.hostgroup.puppet&.all_config_groups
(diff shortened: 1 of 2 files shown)


ezr-ondrej:
CP 1.0: 8693bda

PR #217, opened 2021-10-31
Fixes Puppet params for new Host

ezr-ondrej:
Do not fail to load Host Puppet params for new Host/group


ezr-ondrej:
This is happening only in some occasions and mostly not at all, because the environment is selected thus the puppet is build

ezr-ondrej:
This is mostly about ugly loader being present there and logging 500 ISE, but that should most likely be resolved by "choose env first pls" notice as it is with conf groups

ezr-ondrej:
Closing this in favor of a better/proper fix by #221

PR #222, opened 2021-11-10
i18n - pulling from tx

nadjaheitmann:
https://community.theforeman.org/t/plugin-maintainers-action-required-request-to-extract-strings-for-foreman-3-1/25913

### DIFF
--- a/locale/en/foreman_puppet.po
+++ b/locale/en/foreman_puppet.po
@@ -21,6 +21,9 @@ msgstr ""
 msgid "%s is not in environment"
 msgstr ""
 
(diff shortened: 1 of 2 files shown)


nadjaheitmann:
It did not change the .mo file, is that correct?

ezr-ondrej (approved):
This is updating the PO files which are sources for transifex, we should do it more regularly so :+1: 
Tho to download the translated strings we need to https://github.com/theforeman/foreman/blob/develop/developer_docs/how_to_create_a_plugin.asciidoc#pulling-translations-from-transifex

PR #223, opened 2021-11-11
Skip ConfigGroup migration for missing Hostgroups

ezr-ondrej:
For some reason there are ConfigGroups in database for Hostgroups that were deleted.

### DIFF
--- a/db/migrate/20200803113903_migrate_host_type_in_host_config_groups.foreman_puppet.rb
+++ b/db/migrate/20200803113903_migrate_host_type_in_host_config_groups.foreman_puppet.rb
@@ -13,26 +13,8 @@ class MigrateHostTypeInHostConfigGroups < ActiveRecord::Migration[6.0]
   end
 
   def up
-    host_config_group_ids = FakeHostConfigGroup.for_host.pluck(:host_id).uniq
-    host_config_group_ids.each do |host_id|
-      host_facet_id = ForemanPuppet::HostPuppetFacet.where(host_id: host_id).pick(:id)
-      host_groups = FakeHostConfigGroup.for_host.where(host_id: host_id)
-      if host_facet_id
-        host_groups.update_all(host_type: 'ForemanPuppet::HostPuppetFacet', host_id: host_facet_id)
-      else
-        deleted_groups = FakeConfigGroup.where(id: host_groups.pluck(:config_group_id)).pluck(:name)
-        say "deleting groups #{deleted_groups.join(', ')} from Host (id=#{host_id}) because it has no environment"
-        host_groups.delete_all
-      end
-    end
-    hostgroup_config_group_ids = FakeHostConfigGroup.for_hostgroup.pluck(:host_id).uniq
-    hostgroup_config_group_ids.each do |hostgroup_id|
-      hostgroup_facet_id = ForemanPuppet::HostgroupPuppetFacet.where(hostgroup_id: hostgroup_id).pick(:id)
-      hostgroup_facet_id ||= ForemanPuppet::HostgroupPuppetFacet.create!(hostgroup: Hostgroup.unscoped.find(hostgroup_id)).id
-      FakeHostConfigGroup.for_hostgroup
-                         .where(host_id: hostgroup_id)
-                         .update_all(host_type: 'ForemanPuppet::HostgroupPuppetFacet', host_id: hostgroup_facet_id)
-    end
+    migrate_host_data_to_facets
+    migrate_hostgroup_data_to_facets
   end
 
   def down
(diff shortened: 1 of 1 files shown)


db/migrate/20200803113903_migrate_host_type_in_host_config_groups.foreman_puppet.rb
```diff
@@ -51,4 +33,38 @@ def down
...
+        host_groups.delete_all
+      end
+    end
+  end
+
+  def migrate_hostgroup_data_to_facets
+    hostgroup_config_group_ids = FakeHostConfigGroup.for_hostgroup.pluck(:host_id).uniq
+    hostgroup_config_group_ids.each do |hostgroup_id|
+      hostgroup = Hostgroup.unscoped.find_by(id: hostgroup_id)
+      hostgroup_groups = FakeHostConfigGroup.for_hostgroup.where(host_id: hostgroup_id)
```
nadjaheitmann, line 58: Why is it `host_id: hostgroup_id`?

db/migrate/20200803113903_migrate_host_type_in_host_config_groups.foreman_puppet.rb
```diff
@@ -51,4 +33,38 @@ def down
...
+      if hostgroup
+        hostgroup_facet_id = ForemanPuppet::HostgroupPuppetFacet.where(hostgroup_id: hostgroup_id).pick(:id)
+        hostgroup_facet_id ||= ForemanPuppet::HostgroupPuppetFacet.create!(hostgroup: hostgroup).id
+        hostgroup_groups.update_all(host_type: 'ForemanPuppet::HostgroupPuppetFacet', host_id: hostgroup_facet_id)
```
nadjaheitmann, line 62: Same as above: `host_id: hostgroup_facet_id`

db/migrate/20200803113903_migrate_host_type_in_host_config_groups.foreman_puppet.rb
ezr-ondrej: Because it's actually polymorphic association even though it's named host. I know confusing, but the more interesting part is, that for classes we have separate tables, so we does it differently for each model 🎉  but it's what we're working with coming from core. I'm in favor of changing it, but only after 2.0 is out :)

db/migrate/20200803113903_migrate_host_type_in_host_config_groups.foreman_puppet.rb
```diff
@@ -51,4 +33,38 @@ def down
...
+      else
+        deleted_groups = FakeConfigGroup.where(id: hostgroup_groups.pluck(:config_group_id)).pluck(:name)
+        say "deleting groups #{deleted_groups.join(', ')} from Hostgroup (id=#{hostgroup_id}) because Hosgroup does not exists"
```
nadjaheitmann, line 65:
```suggestion
        say "deleting groups #{deleted_groups.join(', ')} from Hostgroup (id=#{hostgroup_id}) because Hostgroup does not exists"
```

db/migrate/20200803113903_migrate_host_type_in_host_config_groups.foreman_puppet.rb
```diff
@@ -51,4 +33,38 @@ def down
...
+  def migrate_host_data_to_facets
+    host_config_group_ids = FakeHostConfigGroup.for_host.pluck(:host_id).uniq
+    host_config_group_ids.each do |host_id|
+      host_facet_id = ForemanPuppet::HostPuppetFacet.where(host_id: host_id).pick(:id)
+      host_groups = FakeHostConfigGroup.for_host.where(host_id: host_id)
+      if host_facet_id
+        host_groups.update_all(host_type: 'ForemanPuppet::HostPuppetFacet', host_id: host_facet_id)
+      else
+        deleted_groups = FakeConfigGroup.where(id: host_groups.pluck(:config_group_id)).pluck(:name)
+        say "deleting groups #{deleted_groups.join(', ')} from Host (id=#{host_id}) because it doesn't exists or has no environment"
```
nadjaheitmann, line 48:
```suggestion
        say "Deleting config groups #{deleted_groups.join(', ')} from Host (id=#{host_id}) because it does not exist."
```

nadjaheitmann: You don't check for environment, do you? And what does `say` do? Why not using a logger? The string is not marked for translation, either.

db/migrate/20200803113903_migrate_host_type_in_host_config_groups.foreman_puppet.rb
```diff
@@ -51,4 +33,38 @@ def down
...
+        host_groups.delete_all
```
nadjaheitmann, line 49: What if a config group belongs to multiple hosts of which one was deleted? Then it should remain. Does this hold true?

db/migrate/20200803113903_migrate_host_type_in_host_config_groups.foreman_puppet.rb
ezr-ondrej: Say is used in migrations and uses the migration logger. It's a system log so I would not translate.

db/migrate/20200803113903_migrate_host_type_in_host_config_groups.foreman_puppet.rb
ezr-ondrej: Yes, this will run only the SQL, so it will delete nothing else then the joining rows from the join table. Config group will be always kept.

db/migrate/20200803113903_migrate_host_type_in_host_config_groups.foreman_puppet.rb
ezr-ondrej: What to check environment for? :thinking:

ezr-ondrej:
CP 1.0: 426182b

PR #225, opened 2021-11-11
Drop direct references for Puppetclasses

ezr-ondrej:
We are using references through Facets in the plugin, but the old references to Host and Hostgroup were still in place.
We were preventing deletes for users who migrated as those columns were never cleaned up.
We needed these during the migration as old code relied on thos, but now we can safely remove them.


ezr-ondrej:
Merging for 2.0 only

PR #228, opened 2021-11-11
i18n - pulling from tx

nadjaheitmann:


ezr-ondrej (approved):
This at least creates the locale folders, no idea why the translations are empty tho.

ezr-ondrej (approved):
Thanks @nadjaheitmann !

PR #230, opened 2021-11-12
Cleanup environment columns from core tables

ezr-ondrej:
We have left the core tables to include the information.
This fixes it for new plugin installs.


ezr-ondrej:
Terrible ideal @ezr-ondrej ! Please do it properly in separate migration so even users who already installed the plugin would get the fix!

ezr-ondrej:
Ok, I did want to fix the earlier migration where it would make sense, but did not realize we cant fix it once some users already migrated, sorry. Fixed.

ezr-ondrej:
CP 1.0: 382f8b6

PR #233, opened 2021-12-06
[WIP] Environment in taxonomy ignore types

stejskalleos:


ezr-ondrej:
This will need a little bit of patching of the `TaxHost` as
1. it can't deal with the namespaced ignore_types properly (it will derive a key `'foreman_puppet/environment_ids'`)
2. it doesn't have `environment_id` in the `HASH_KEYS` so it would not do the main operations anyway

I've started with https://github.com/theforeman/foreman_puppet/pull/119 tho we should have a discussion if it is necessary, or we just want to migrate `ingnore_types: [Environment]` to `ForemanPuppet::Environment.each { |env| <add env to taxonomy> through taxable_taxonomies }` and do not add this in the ignore types at all. Note that it would be missing from mismatched report, which is kinda bad (you'll no longer see that the Environment is not in a Taxonomy you are trying to put the Host into and so on)

ezr-ondrej:
Thinking about this we should get this going and we won't be able to finish off #119.

@stejskalleos do you think you could make the migration for all Taxonomies that have `ignore_types.include?('Environment')` to create TaxableTaxonomy records per each environments? The meaning would be similar in current instances and we would fix the failures. The regression in the funcionality should be tackled in #119.

ezr-ondrej:
Closing in favor of #256

PR #236, opened 2021-12-10
Remove content_tag_for helper

stejskalleos:
For https://projects.theforeman.org/issues/28571


app/views/foreman_puppet/config_groups/_config_group.html.erb
```diff
@@ -1,4 +1,8 @@
-<%= content_tag_for :li, config_group, (added ? :selected : nil), :id => "config_group_#{config_group.id}", :class => css_class.to_s do %>
+<%
+  li_id = "#{added ? 'selected_' : '' }config_group_#{config_group.id}"
+  li_class = "#{added ? 'selected_config_group' : 'config_group'} #{css_class}"
+%>
+<li id="<%= li_id %>" class="<%= li_class %>">
```
ekohl, line 5: Could this use regular `content_tag` instead?

ezr-ondrej (approved):
We probably could have inline id and class, but I'm up for this, might even look tidier :+1:
Thanks @stejskalleos !

PR #238, opened 2021-12-12
Drop SETTINGS[:unattended] check

### LINKED ISSUE #237: Drop unattended: false mode
This has been deprecated in core, and we shouldn't support them in the puppet plugin either.

tbrisker:
It's deprecated in core.

Fixes #237.


ezr-ondrej (approved):
Thanks @tbrisker ! :+1:

PR #241, opened 2021-12-17
Upgrade compatibility matrix in README

nadjaheitmann:
I have seen multiple people on discourse who have the wrong Foreman/plugin combinations. Do we need to adjust packaging or something like that as well? Or Foreman dependency in our gemspec?

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ You can install it on Foreman 2.5 to prepare for the Foreman update.
 
 |Foreman version|Plugin version|Notes                                     |
 |---------------|--------------|------------------------------------------|
-| >= 3.0        | ~> 1.0       | Required                                 |
+| >= 3.2        | ~> 3.0       | Required                                 |
+| ~> 3.1        | ~> 2.0       | Required                                 |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,7 +37,9 @@ You can install it on Foreman 2.5 to prepare for the Foreman update.
 
 |Foreman version|Plugin version|Notes                                     |
 |---------------|--------------|------------------------------------------|
-| >= 3.0        | ~> 1.0       | Required                                 |
+| >= 3.2        | ~> 3.0       | Required                                 |
```
ezr-ondrej, line 40: not sure if we won't have 2.1 instead (unless we have breaking changes.
But I guess it doesn't matter, we can just release 3.0 :shrug:

ezr-ondrej (approved):
Thanks @nadjaheitmann

README.md
nadjaheitmann: I kind of assumed it will be 3.0 as there is a 3.0 tag already. In any case, we can adjust the matrix when we get there.

PR #249, opened 2022-02-01
Fixes tests after core changes

### LINKED ISSUE #248: Remove tests for require_ssl_smart_proxies
This setting has been droped in https://projects.theforeman.org/issues/34236

ezr-ondrej:
The setting has been dropped, so we should not test for it.

Fixes #248

Needs https://github.com/theforeman/foreman/pull/9081 for some tests


ezr-ondrej:
It resolves the test failures :shrug:

PR #250, opened 2022-02-01
Hide Puppet params table until environment is selected

### LINKED ISSUE #221: Show notice until Environment is chosen instead of params table
We show params table even when Environment is not chosen.
We should just show the notice urging user to chose Environment first.

ezr-ondrej:
Fixes #221

### DIFF
--- a/app/views/hosts/_form_puppet_enc_tab.html.erb
+++ b/app/views/hosts/_form_puppet_enc_tab.html.erb
@@ -11,13 +11,14 @@
   <% end %>
 
   <%= render 'foreman_puppet/puppetclasses/class_selection', obj: obj, resource_type: resource_type %>
+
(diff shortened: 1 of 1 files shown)


nadjaheitmann:
If you remove the environment, the selection option for the classes remains. Not sure if this is much of a problem.

ezr-ondrej:
> the selection option for the classes remains

That is little weird and would be nice to fix, but I'd consider it not part of this issue?

nadjaheitmann:
> > the selection option for the classes remains
> 
> That is little weird and would be nice to fix, but I'd consider it not part of this issue?

Well, technically, it is, because if you deselect the environment and you still have a parameter selection, this is the same issue as before. However, I would assume that no one disselects an environment after an initial selection, so I would go ahead and merge.

nadjaheitmann (approved):

PR #251, opened 2022-02-01
Add forgotten permitted action for edit_classes

### LINKED ISSUE #244: "Organization admin" role doesn’t have permission to edit host’s Puppet classes
<div class="cooked"><p><strong>Problem:</strong><br>
User assigned with “Organization admin” role doesn’t have permission to edit Puppet classes associated with specific host.</p>
<p>Logged in as user with “Organization admin” role, navigate to Hosts 
&gt; [example_host] &gt; Edit &gt; Puppet ENC. Adding new or removing 
existing Puppet classes isn’t possible. The following message appears 
when hovering over classes: “Not authorized to edit classes”</p>
<p>Adding and removing config groups works fine.</p>
<p><strong>Expected outcome:</strong><br>
The user shouldn’t receive the “Not authorized to edit classes” message and it should be able to add/remove Puppet classes.</p>
<p><strong>Foreman and Proxy versions:</strong><br>
3.1.0</p>
<p><strong>Foreman and Proxy plugin versions:</strong></p>
<div class="md-table">

Plugin | Version
-- | --
foreman_memcache | 0.1.1
foreman_puppet | 2.0.0
foreman_templates | 9.1.0
foreman_vault | 1.1.0
puppetdb_foreman | 5.0.0

</div><p><strong>Distribution and version:</strong><br>
CentOS Linux release 7.9.2009 (Core)</p>

**Other relevant data:**
“Organization admin” role is assigned with `edit_classes` filter for HostClass resource type, which should allow adding/removing of Puppet 
classes.</p></div>

The user is associated with specific organization and location (the same organization and location in which the host is “located”).

I'm experiencing the same problem if the user is assigned with the "Manager" role.

ezr-ondrej:
We have removed the action from edit_classes permission,
but even tho the controller does not exists, we are checking for permissions on it.

This adds the permitted action of host_editing/edit_classes to edit_classes permission.

Fixes #244

### DIFF
--- a/db/migrate/20220201205305_migrate_host_class_permission.foreman_puppet.rb
+++ b/db/migrate/20220201205305_migrate_host_class_permission.foreman_puppet.rb
@@ -1,0 +1,9 @@
+class MigrateHostClassPermission < ActiveRecord::Migration[6.0]
+  def up
+    Permission.where(resource_type: 'HostClass').update_all(resource_type: 'ForemanPuppet::HostClass')
+  end
+
+  def down
+    # no can do
+  end
+end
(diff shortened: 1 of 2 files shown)


nadjaheitmann (approved):
Seems that the problem is fixed for the user who reported it. Should be good then.

ezr-ondrej:
CP 2.0: f80660a

PR #259, opened 2022-03-01
Correctly show help buttons

ezr-ondrej:
Core helper is designed to work only for core controllers.
The easiest way to fix the buttons is to not use the helper as it does very little.


ezr-ondrej:
CP 2.0: 507416c

PR #260, opened 2022-03-03
Remove test that doesnt work

ezr-ondrej:


nadjaheitmann (approved):
Just go for it if the test is not useful.

ezr-ondrej:
> Just go for it if the test is not useful.

It doesnt test what it is supposed to anymore :shrug:
And it would be too hard to test properly and it should be tested in core

PR #262, opened 2022-03-14
Removes hack for Site manager role

ezr-ondrej:
We've introduced a hack for Site manager role to be able to add perms to it.
In https://projects.theforeman.org/issues/34351 we have implemented the proper solution tho.
Since that we do not need this workaround.


ezr-ondrej:
CP 3.0: 55ddcb3

nadjaheitmann:
CP 2.0: 94c1dbe68956cc02409d295e9517d53da5862c97

PR #261, opened 2022-03-10
Add details card to reports sub tab.

amirfefer:

PR #274, opened 2022-05-07
Add details card

Ron-Lavi:
Continues @amirfefer work from #261
cc @nadjaheitmann @marisvirik

Screenshot (82)

Screenshot (83)

### DIFF
--- a/webpack/src/Extends/Host/PuppetTab/Routes.js
+++ b/webpack/src/Extends/Host/PuppetTab/Routes.js
@@ -5,11 +5,11 @@ import { route } from './helpers';
 import EmptyPage from './SubTabs/EmptyPage';
 import Reports from './SubTabs/Reports';
 
-const SecondaryTabRoutes = ({ hostName }) => (
+const SecondaryTabRoutes = ({ hostName, hostInfo, status }) => (
   <Switch>
     <Route path={route('reports')}>
       {hostName ? (
-        <Reports hostName={hostName} />
+        <Reports hostName={hostName} hostInfo={hostInfo} status={status} />
       ) : (
         <EmptyPage header="Reports" />
       )}
@@ -29,10 +29,14 @@ const SecondaryTabRoutes = ({ hostName }) => (
 
 SecondaryTabRoutes.propTypes = {
   hostName: PropTypes.string,
+  hostInfo: PropTypes.object,
+  status: PropTypes.string,
 };
 
 SecondaryTabRoutes.defaultProps = {
   hostName: '',
+  hostInfo: {},
+  status: undefined,
 };
 
 export default SecondaryTabRoutes;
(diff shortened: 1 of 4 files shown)


Ron-Lavi:
for the upcoming release, I don't think we will manage to work on the `Last config status` card,
should we remove it for now? or keep it with `TBD`?

MariSvirik:
I vote to remove it. It was designed to use a global status, anyway.

Ron-Lavi:
Removed that card and shifted the Puppet details card to the left for now
Screenshot (85)

nadjaheitmann:
@Ron-Lavi  Looks good to me. What is the policy about unit testing for React?

Ron-Lavi:
@nadjaheitmann we should probably add full rails integration test to see if the tab is created with some reports data,
I will try adding something into: [test/integration/foreman_puppet/host_js_test.rb](https://github.com/theforeman/foreman_puppet/blob/master/test/integration/foreman_puppet/host_js_test.rb)

nadjaheitmann:
I like this log message in the tests:
`Skipped:
ForemanPuppet::HostTest::#info puppet bits#test_0002_ENC YAML uses Classification::ClassParam for parameterized output [/home/runner/work/foreman_puppet/foreman_puppet/foreman_puppet/test/models/foreman_puppet/host_test.rb:120]:
No idea whats wrong here`
(not related to your test, but it looks like the new test is failing)

test/integration/foreman_puppet/host_js_test.rb
```diff
@@ -101,6 +101,15 @@ class HostJSTest < IntegrationTestWithJavascript
...
         assert_equal env.name, host.puppet.environment.name
       end
+
+      test 'puppet tab' do
+        host = FactoryBot.create(:host)
+
+        visit host_details_page_path(host)
+
+        find('button', text: 'Puppet').click
+        assert find('section > div.report-tab > div.pf-c-card__title').has_text? 'Puppet details'
```
Ron-Lavi, line 111: meanwhile, I see that the there's an issue with how the test environment loads the plugin's bundled JS,
I see that the plugin JS file is loaded in the new host page `head`, however the slot and fill doesn't seem to work during the test and I see the browser's console error: with the error: 
`Uncaught TypeError: Cannot read properties of undefined (reading 'call')` somewhere from webpack. 
I didn't see any other integration tests in other plugins, though we should fix it, probably not in this PR,
I would suggest to make it in a follow up

Ron-Lavi: cc @nadjaheitmann @MariaAga @amirfefer

nadjaheitmann: Sounds good.

nadjaheitmann:
@Ron-Lavi Can you just confirm once again, this is ready to go now, right?

Ron-Lavi:
Yes @nadjaheitmann thanks!

nadjaheitmann:
Thank you @Ron-Lavi !

nadjaheitmann:
Any ideas why the tests are failing in master? Wasn't it working just yesterday?

MariaAga:
If you mean the J's tests, there is an issue with JavaScript tests, and will be fixed in foreman-js.
Caused by new @babel version 7.17.12.

PR #281, opened 2022-05-17
Remove ruby spec tests

Ron-Lavi:
@nadjaheitmann let's see if it fixes the ruby tests CI


ezr-ondrej:
It's just the specs testing which I'm in favor of removing as I'm the one who pushed for those, but I'll not be maintaining the plugin.
So unless someone else see the value there, lets get rid of them completely

nadjaheitmann:
I am not familiar with spec tests, to be honest. So @Ron-Lavi, @amirfefer or @MariaAga, if you think it is worth keeping the tests, let's keep them. Otherwise, let's remove them.

PR #284, opened 2022-05-31
Fixes #283 - Re-add permissions for smart proxy class import

### LINKED ISSUE #283: Permissions for importing classes from smart proxies missing
https://community.theforeman.org/t/import-puppetclass-permissions-no-longer-work/28821

nadjaheitmann:


ares (approved):
Makes sense, thanks for the fix!

Manisha15:
> This patch appears to have broken the API on our Foreman instance when trying to list hosts.
> 
> ```
> {
>   "error": {"message":"Cannot find rabl template 'foreman_puppet/api/v2/host_puppet_facets/host_list' within registered ([\"/usr/share/foreman/app/views\"]) view paths!"}
> }
> ```
> 
> I am pretty sure this change should be `foreman_puppet/api/v2/smart_proxies` instead of just `api/v2/smart_proxies`.
> 
> Adding the prefix manually in the filesystem on our Foreman server and restarting Foreman fixes the problem.
> 
> Can we get a fix and a release as soon as possible?

Is this for a particular `foreman` version or all of them?

nadjaheitmann:
The api prefix was removed from the API on purpose, see https://github.com/theforeman/foreman_puppet/commit/5ab36aa16d497d2c04f3968b1d817eddac784c78 .
@ncstate-mafields Is your apipie-cache up-to-date? E.g. did you execute `foreman-rake apipie:cache` on your system?

PR #285, opened 2022-05-31
Update readme

nadjaheitmann:


nadjaheitmann:
@kamils-iRonin Hi Kamil, you have implemented graphql API for foreman_puppet, right? The tests are giving the following warning:
`/home/runner/work/foreman_puppet/foreman_puppet/vendor/bundle/ruby/2.7.0/gems/graphql-1.8.18/lib/graphql/execution/multiplex.rb:85: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call`

Are there fixes required?

ezr-ondrej:
@nadjaheitmann @kamils-iRonin indeed, it is basicaly about  using obsolete graphql and indeed it will need to be bumped to upgrade to ruby 3.0 and therefore the warning
see https://projects.theforeman.org/issues/32685 for the first take on bump, but it did fail because we are using concepts that are no longer supported in new version of the gem, while it is not THAT hard to update, it will require some work. To anyone who would decide to take that on, I'd be happy to relay any info I remember when I was digging in GraphQL a bit.

nadjaheitmann:
Thanks for the info. I will open an issue for that.

nadjaheitmann:
Oh, I see that @kamils-iRonin was already asked graphql questions before https://github.com/theforeman/foreman/pull/9230 :D

PR #290, opened 2022-06-13
Remove EmptyPage demo description

Ron-Lavi:
and also refactor component that didn't actually need it.


webpack/src/Extends/Host/PuppetTab/SubTabs/ENCPreview/index.js
```diff
@@ -23,21 +23,20 @@ const ENCPreview = ({ hostName }) => {
     return <Skeleton count={5} />;
   }
 
-  if (status === STATUS.ERROR) {
+  if (status === STATUS.ERROR || !hostName) {
+    const description = !hostName
+      ? `Couldn't find data for host: ${hostName}`
```
Manisha15, line 28: If `hostName` is empty then how is it going to print the first message with `${hostName}` ?

Manisha15 (approved):

PR #305, opened 2022-08-23
Fix the taxable_taxonomies creation

ares:
When the migration creates the records in taxable_taxonomies, it does
not set any value to created_at and updated_at magic columns. It uses
`insert_all` which skips callbacks, so they remain null. There are
databases which enforces the column not to be null, therefore we need to
be setting the values manually.

### DIFF
--- a/db/migrate/20220208135305_migrate_environment_ignore_type.foreman_puppet.rb
+++ b/db/migrate/20220208135305_migrate_environment_ignore_type.foreman_puppet.rb
@@ -7,7 +7,8 @@ class MigrateEnvironmentIgnoreType < ActiveRecord::Migration[6.0]
       new_types = tax.ignore_types.reject { |type| type == 'Environment' }
       tax.update_columns(ignore_types: new_types)
       taxable_rows = environment_ids.map do |env_id|
-        { taxable_id: env_id, taxable_type: 'ForemanPuppet::Environment', taxonomy_id: tax.id }
+        data = { taxable_id: env_id, taxable_type: 'ForemanPuppet::Environment', taxonomy_id: tax.id }
+        TaxableTaxonomy.column_names.include?('created_at') ? data.merge({ created_at: Time.zone.now, updated_at: Time.zone.now }) : data
       end
       TaxableTaxonomy.insert_all(taxable_rows) if taxable_rows.any?
     end


db/migrate/20220208135305_migrate_environment_ignore_type.foreman_puppet.rb
```diff
@@ -7,7 +7,7 @@ def up
       new_types = tax.ignore_types.reject { |type| type == 'Environment' }
       tax.update_columns(ignore_types: new_types)
       taxable_rows = environment_ids.map do |env_id|
-        { taxable_id: env_id, taxable_type: 'ForemanPuppet::Environment', taxonomy_id: tax.id }
+        { taxable_id: env_id, taxable_type: 'ForemanPuppet::Environment', taxonomy_id: tax.id, created_at: Time.now, updated_at: Time.now }
```
ezr-ondrej, line 10: :police_officer: not happy about this

ares:
The test failure seems unrelated to me

```
Error:
Queries::PuppetclassQueryTest#test_0001_fetching puppetclass attributes:
ActiveRecord::UnknownAttributeReference: Query method called with non-attribute argument(s): nil
    app/graphql/collection_loader.rb:44:in `preloader_for_association'
    app/graphql/collection_loader.rb:26:in `perform'
    test/test_helper.rb:187:in `block in <class:GraphQLQueryTestCase>'
    foreman_puppet/test/graphql/queries/puppetclass_query_test.rb:56:in `block in <class:PuppetclassQueryTest>'
```

Manisha15:
> The test failure seems unrelated to me
> 
> ```
> Error:
> Queries::PuppetclassQueryTest#test_0001_fetching puppetclass attributes:
> ActiveRecord::UnknownAttributeReference: Query method called with non-attribute argument(s): nil
>     app/graphql/collection_loader.rb:44:in `preloader_for_association'
>     app/graphql/collection_loader.rb:26:in `perform'
>     test/test_helper.rb:187:in `block in <class:GraphQLQueryTestCase>'
>     foreman_puppet/test/graphql/queries/puppetclass_query_test.rb:56:in `block in <class:PuppetclassQueryTest>'
> ```

Yeah that is failing for all PRs.

Manisha15 (approved):

PR #312, opened 2022-10-28
Fixes host details puppet tab crashing on enter

MariaAga:
No need to use `cardTemplate` as it used more for tabs with a lot of cards that need container logic and remembering if cards are expended or not.

### DIFF
--- a/webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/DescriptionCard.js
+++ b/webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/DescriptionCard.js
@@ -1,7 +1,11 @@
 import PropTypes from 'prop-types';
-import React from 'react';
-import CardTemplate from 'foremanReact/components/HostDetails/Templates/CardItem/CardTemplate';
+import React, { useState } from 'react';
 import {
+  Card,
+  CardTitle,
+  CardHeader,
+  CardBody,
+  CardExpandableContent,
   DescriptionList,
   DescriptionListTerm,
   DescriptionListGroup,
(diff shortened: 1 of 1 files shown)


MariaAga:
test failure unrelated

webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/DescriptionCard.js
```diff
@@ -1,7 +1,11 @@
 import PropTypes from 'prop-types';
-import React from 'react';
-import CardTemplate from 'foremanReact/components/HostDetails/Templates/CardItem/CardTemplate';
```
Ron-Lavi, line 3: Imho it would be better to use `CardTemplate` from core as it is,
but protect plugins in case the [Expansion Context](https://github.com/theforeman/foreman/blob/2c252511913090d3defe58cbc863af5c5920811d/webpack/assets/javascripts/react_app/components/HostDetails/Templates/CardItem/CardTemplate/index.js#L27) is `undefined`

Ron-Lavi (commented):
Thanks for the PR @MariaAga!
Still thinking if this should be resolved in core or plugins,
as I would be happy to receive that component from core.

webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/DescriptionCard.js
MariaAga: I disagree, because cardtemplate needs `dispatch` from context in order to expand, and while its good that the page doesn't crash with the foreman pr the card is not expandable

MariaAga:
Fixed in foreman

PR #313, opened 2022-10-31
Fixes #35697 - Make foreman_puppet use selectable columns

pkoprda:


ares:
Is there something we can do to make it not appear as the first column before the name? It may be due to the fact, it is it's own category, but this should likely not be the first column of the table
Screenshot_20221102_084909

PR #319, opened 2022-12-05
Fixes #35822 - Renaming header of selectable column

pkoprda:
Changing header name from `Environment name` to `Puppet env`.


Ron-Lavi (approved):
Thanks @pkoprda!

PR #322, opened 2022-12-13
GraphQL: Fix graphql tests

kamils-iRonin:


app/models/foreman_puppet/puppetclass.rb
```diff
@@ -15,8 +15,8 @@ class Puppetclass < ApplicationRecord
     before_destroy EnsureNotUsedBy.new(:hosts, :hostgroups)
     has_many :environment_classes, dependent: :destroy, inverse_of: :puppetclass
     has_many :environments, -> { distinct }, through: :environment_classes
-    has_many :organizations, -> { distinct.reorder(nil) }, through: :environments
-    has_many :locations, -> { distinct.reorder(nil) }, through: :environments
```
kamils-iRonin, line 19: this causes `ActiveRecord::UnknownAttributeReference: Query method called with non-attribute argument(s): nil` error

Ron-Lavi (approved):
Thanks @kamils-iRonin !

PR #321, opened 2022-12-13
Fix broken CardTemplate

Ron-Lavi:
Using CardTemplate from Foreman in foreman_puppet v4.0.3, led to the puppet tab not functioning.

PR #326, opened 2022-12-20
Bump foreman_puppet to 4.0.4

Ron-Lavi:


Ron-Lavi:
in continues to the work in https://github.com/theforeman/foreman_puppet/pull/321#issuecomment-1356759364
we will do v4.0.4 with the fix of 4.0-stable
and will also branch right after to 4.1-stable

Manisha15 (approved):

PR #329, opened 2023-01-26
Fixes issue 325 - Correct untranslatable welcome string

### LINKED ISSUE #325: Untranslatable string
This string can't be translated:
https://github.com/theforeman/foreman_puppet/blob/41ab916bbbc78cfdaff8fa7723df88165f31cbda/webpack/src/Components/Environments/Welcome.js#L13-L16

It's **always** invalid to use ```__(````.

_Originally posted by @ekohl in https://github.com/theforeman/foreman_puppet/pull/323#discussion_r1050737749_

nadjaheitmann:
I did the fix based on https://github.com/theforeman/foreman/commit/57a4506c6060dbf34e6b94e322c2316b1d50a808

Please check if there is anything I missed.

### DIFF
--- a/webpack/src/Components/Environments/Welcome.js
+++ b/webpack/src/Components/Environments/Welcome.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
 import { translate as __ } from 'foremanReact/common/I18n';
 import EmptyState from 'foremanReact/components/common/EmptyState';
 import { foremanUrl, getManualURL } from 'foremanReact/common/helpers';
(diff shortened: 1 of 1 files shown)


Ron-Lavi (approved):
Thanks @nadjaheitmann !

webpack/src/Components/Environments/Welcome.js
```diff
@@ -10,16 +11,34 @@ export const WelcomeEnv = ({ canCreate }) => {
...
-  const content = __(`If you are planning to use Foreman as an external node classifier you should provide information about one or more environments.<br/>
-  This information is commonly imported from a pre-existing Puppet configuration by the use of the <a target="_blank" href=${getManualURL(
-    '4.2.2Classes'
-  )}>Puppet classes and environment importer.</a>`);
+  const description = (
+    <>
+      <FormattedMessage
+        id="puppetenv-welcome"
+        defaultMessage={__(
+          'If you are planning to use Foreman as an external node classifier you should provide information about one or more environments.{newLine}This information is commonly imported from a pre-existing Puppet configuration by the use of the {puppetClasses}.'
```
ekohl, line 19: Late to the party, so sorry about that.

You have improved it, but you still have made something that's very hard to translate because you string parts of a sentence together. In a language where the word order is not the same it may be incomprehensible for the reader.

What I think you should probably do is have a `{link-start}` and `{link-end}` placeholder so translators have the freedom to move it around.

ekohl: Oh, and sadly we can't extract multi-line strings now so this string doesn't show up in the .pot file.

nadjaheitmann:
Thanks @ekohl for your comments. It would be nice to have those information in the translations guide: https://projects.theforeman.org/projects/foreman/wiki/Translating

ekohl:
Yes, I'm slowly collecting my thoughts on that, but it's also taking me a while to grasp. Right now I'm collecting the issues and seeing which ones we should be solving at the core (like the multi-line string extraction) and the ones you as the author must be aware of (like stringing together sentences). Thanks for sticking with it.

PR #335, opened 2023-02-17
Remove unused rspec-rails dependency

ekohl:
The rspec tests were dropped, so the dependency can also be dropped.

Fixes: 958841594157ec52e5d8c3abc40bbefdad3ef953

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,5 +19,4 @@ end
 
 gemspec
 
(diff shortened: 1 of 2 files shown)


kamils-iRonin (approved):

kamils-iRonin:
Thanks @ekohl !

Issue #315, opened 2022-11-04
Searching with `puppetclass ~ <class>` causes undefined method `allows_taxonomy_filtering?' for ForemanPuppet::Extensions::Host:Module


laugmanuel:
When searching for specific puppetclasses, the UI reports:
```sh
external method 'search_by_puppetclass' failed with error: undefined method `allows_taxonomy_filtering?' for ForemanPuppet::Extensions::Host:Module
```

Seems like the issue only affects non-admin users.

Manisha15:
Which foreman_puppet version are you using?

laugmanuel:
@Manisha15 can confirm that #341 fixes this issue.

nadjaheitmann:
Fixed by #341

PR #341, opened 2023-02-27
Fixes Host view permissions to search by puppetclass

Manisha15:


Manisha15:
Fixes issue #315

nadjaheitmann (commented):
I am not exactly sure what the code is supposed to do, but at the first glance it looks like the code prevents users to access hosts that they are not authorized to access. If so, I suppose that this restriction should not be removed.

Manisha15:
> I am not exactly sure what the code is supposed to do, but at the first glance it looks like the code prevents users to access hosts that they are not authorized to access. If so, I suppose that this restriction should not be removed.

I have reverted the authorization permissions and now this should work. We should also add the documentation somewhere that `search by puppetclass` is only available for users with `view_hosts` role.

Ron-Lavi:
@laugmanuel can I ask you to verify the fix again?

laugmanuel:
@Ron-Lavi still works!

nadjaheitmann:
Merged, thanks @Manisha15 !

nadjaheitmann:
CP to 4.1: ab289f86b14ea266e5c74750fddc4d7866a4b0f7

PR #344, opened 2023-03-01
Fixes #35949 - move select_multiple_puppet_proxy to the plugin

Ron-Lavi:
Corresponding PR in https://github.com/theforeman/foreman/pull/9636

### DIFF
--- a/app/views/hosts/select_multiple_puppet_proxy.html.erb
+++ b/app/views/hosts/select_multiple_puppet_proxy.html.erb
@@ -1,0 +1,5 @@
+<%= render 'selected_hosts', :hosts => @hosts %>
+
+<%= form_for :proxy, :url => foreman_puppet.update_multiple_puppet_proxy_hosts_path(:host_ids => params[:host_ids]) do |f| %>
+  <%= multiple_proxy_select(f, 'Puppet') %>
+<% end %>


nadjaheitmann (approved):
I patched a test system and it seems to fix the isse.

Ron-Lavi:
> I patched a test system and it seems to fix the issue.

Hey @nadjaheitmann, even without the corresponding Foreman PR?

nadjaheitmann:
> > I patched a test system and it seems to fix the issue.
> 
> Hey @nadjaheitmann, even without the corresponding Foreman PR?

I tested with the corresponding Foreman PR, I did not without.

nadjaheitmann:
> I tested with the corresponding Foreman PR, I did not without.

Does not work without the change in Foreman ;)

nadjaheitmann:
> > @laugmanuel Could you also verify, please?
> 
> Unfortunately, I don't have a somewhat "new" and working Foreman instance I could use for testing - sorry :(

Okay. But the issue should also affect Foreman 3.1, shouldn't it? In case you need the fix.

nadjaheitmann:
Shall we go ahead and merge it? Even if the patch is not applied on Foreman, the change does not break anything.

laugmanuel:
> Okay. But the issue should also affect Foreman 3.1, shouldn't it? In case you need the fix.

Oh you are completely right! Thank's for the heads-up. It's very uncommon for us to use that feature and therefore we did not even notice.  
I've applied both patches and everything looks fine afterwards!

laugmanuel (approved):

PR #349, opened 2023-05-15
Translation improvements

stejskalleos:
* Extract missing strings for the translations
* Update config for the transifex-cli
* Update *.po files
* Fix interpolations in strings


nadjaheitmann:
@stejskalleos I thought it might be a good idea to have some of the bug fixes available for previous versions before we need to bump a new major version. So I just released a new minor version, just in case you're wondering what happened :)

stejskalleos:
Fixed & it's green

nadjaheitmann (approved):

Issue #307, opened 2022-09-14
Smart Proxies → <Foreman-system> → Puppet shows the Distribution Chart in a broken way


Ron-Lavi:
From https://community.theforeman.org/t/notable-issues-in-foreman-3-4-katello-4-6/30222
If you go to Infrastructure → Smart Proxies → <Foreman-system> → Puppet, it will show the Distribution Chart in a broken way (after reloading the page with i.e. F5 if will show it correctly)

![a6228ad25f6f91f6541bc0b13e0163edb4e02d0a_2_1034x662](https://user-images.githubusercontent.com/26363699/190143453-7960fe7c-264b-4c8c-b648-8b955dbd2a76.png)

stejskalleos:
@nadjaheitmann Is this still relevant, or can it be closed?

nadjaheitmann:
Issue is still persisting.

PR #352, opened 2023-05-22
Fixes #36421 - do not load non-existant charts.js

evgeni:
We used charts.js from Foreman, but that was dropped in Foreman 3.0 (https://github.com/theforeman/foreman/commit/69379b51a33344de93836950565ecbfae0c4df23) and results in 404 errors when trying to load the smart proxy details page.

### DIFF
--- a/app/views/smart_proxies/plugins/_puppet.html.erb
+++ b/app/views/smart_proxies/plugins/_puppet.html.erb
@@ -1,4 +1,3 @@
-<% javascript 'charts' %>
 <ul id="proxy-puppet-tab" class="nav nav-tabs nav-tabs-pf">
   <li class="active"><a href="#puppet_dashboard" data-toggle="tab"><%= _('General')%></a></li>
   <li><a href="#puppet_envs" data-toggle="tab"><%= _('Environments') %></a></li>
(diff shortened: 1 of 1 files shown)


evgeni:
Huh. Seems there actually is a broken chart (not that it was working before I removed the loading of the non-existant file, but…)

app/views/smart_proxies/plugins/_puppet.html.erb
```diff
@@ -16,7 +15,7 @@
         </div>
       </div>
     </div>
-    <div class="clearfix row container-cards-pf container-fluid" data-ajax-url="<%= foreman_puppet.dashboard_puppet_smart_proxy_path(@smart_proxy) %>" data-on-complete="refreshCharts"><%= spinner %></div>
```
evgeni, line 19: this function came from charts.js :woman_shrugging:

nadjaheitmann:
Thanks @evgeni . Does this fix https://github.com/theforeman/foreman_puppet/issues/307 ? 
And @stejskalleos tests are failing with `ActionView::Template::Error: DEPRECATION WARNING: The asset "foreman_puppet/locale/en/foreman_puppet.js" is not present in the asset pipeline.` Is this related to the translations update?

evgeni:
No, this should have no effect on 307.

stejskalleos (changes requested):
https://github.com/theforeman/foreman_puppet/pull/353 merged, please rebase so we can have green tests

PR #358, opened 2023-06-17
Deal no selected environment in host form

ekohl:
If you create a host without a Puppet environment then the Puppet ENC tab is empty. This means the puppet_klasses_parameters element isn't in the DOM. This means url ends up being undefined. In that case the Puppet class parameters can't be refreshed.

I don't know if this is correct, but we just ran into this. It breaks the media selection in Katello.

cc @ares


nadjaheitmann:
Just had a look. This solves the issue you describe and I did not see any other issues during testing.

nadjaheitmann (approved):

nadjaheitmann:
CP 5.0-stable: cb2fa5c2bdc85c258d9e708ed48bc9ab4910917d

PR #372, opened 2023-11-14
fix rendering of the react layout

evgeni:
it's currently unused, so nobody reported the issues yet


stejskalleos (approved):
:green_apple: **LGTM**

PR #376, opened 2023-11-23
Add missing [Change Puppet CA] action

stejskalleos:


ekohl (commented):
It's a bit confusing, but the Puppet CA bit is in Foreman core. So there is for example this bit file: https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/foreman/controller/puppet/hosts_controller_extensions.rb

Shouldn't it actually be a patch in core then? Or in other words, (partially) revert https://github.com/theforeman/foreman/commit/6cdb8c7a9ebd790537510213e43537a4a87189d6

I left it as a comment: https://github.com/theforeman/foreman/pull/9636/files#r1403383879

stejskalleos:
I guess so, I didn't know that Puppet CA is in the core ... Closing the BZ, I'll fix it in the core.

PR #377, opened 2023-11-24
Prevent Puppet subtabs from disappearing on scrolling

nadjaheitmann:
@MariaAga Could you have a look, please?

PR #379, opened 2023-12-01
Implement last config status tab

nadjaheitmann:
The first commit of this PR is part of #377


nadjaheitmann:
image

webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/ConfigStatusCard.js
```diff
@@ -0,0 +1,179 @@
...
+          metrics.corrective_change
+        )}
+      </Tr>
+      <Tr>
+        {createPuppetMetricsTableElement(__('Skipped'), metrics.skipped)}
+        {createPuppetMetricsTableElement(
+          __('Out of sync'),
+          metrics.out_of_sync
+        )}
+        {createPuppetMetricsTableElement(__('Total'), metrics.total)}
```
nadjaheitmann, line 120: Actually, those are very close to the actual Puppet metrics name, so maybe we should drop translations here?

stejskalleos:
@MariaAga could you take a look at the JS, please?

webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/ConfigStatusCard.js
```diff
@@ -0,0 +1,179 @@
...
+    <>
+      {__('Last configuration status')}
+      <Flex>
+        <FlexItem>
+          <Button
+            ouiaId="foreman-puppet-last-report-button"
+            variant="link"
+            component="a"
+            isInline
+            isDisabled={allReports === []}
```
MariaAga, line 52: This is always false, please use `!all.reports.length`

webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/ConfigStatusCard.js
```diff
@@ -0,0 +1,179 @@
...
+  const dispatch = useDispatch();
+  // get already fetched results from reports tab
+  const API_KEY = `get-reports-${hostName}`;
+  const { reports, itemCount } = useSelector(state =>
+    selectAPIResponse(state, API_KEY)
+  );
+
+  // we need to fetch the last Puppet report to get all Puppet metrics
+  const getLastReport = useCallback(() => {
+    if (hostName && reports) dispatch(getReportByIdAction(reports[0].id));
```
MariaAga, line 136: empty objects are still true in JS, please use `reports?.length`

webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/ConfigStatusCard.js
```diff
@@ -0,0 +1,179 @@
...
+    getLastReport();
+  }, [hostName, reports]);
+
+  const { metrics } = useSelector(state =>
+    selectAPIResponse(state, FOREMAN_PUPPET_LAST_REPORT_KEY)
+  );
+
+  return (
+    <CardTemplate
+      header={generateCardHeader(hostName, reports, itemCount)}
```
MariaAga, line 149: hostName is never used in generateCardHeader

webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/ConfigStatusCard.js
```diff
@@ -0,0 +1,179 @@
...
+    <> {__('No configuration status available')} </>
+  );
+
+const createPuppetMetricsTableElement = (name, value) => (
+  <>
+    <Td modifier="truncate" key={`metrics-name-${name}`}>
+      <TableText>{name}</TableText>
+    </Td>
+    <Td modifier="truncate" key={`metrics-name-${value}`}>
+      <TableText>{value}</TableText>
```
MariaAga, line 84: I sometimes see 0, and sometimes no value. can we have something instead of an empty space when there is no value? 
image

MariaAga (changes requested):

webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/ConfigStatusCard.js
nadjaheitmann: Any preferences about the default value?

nadjaheitmann (commented):
Thanks @MariaAga !!

webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/ConfigStatusCard.js
MariaAga: Does "not applicable" (`n/a`) represent correctly what it means when a metric is null?

nadjaheitmann: Personally, I don't know if the value from the Puppet report can be empty at all. But for me, (n/a) sounds reasonable.

nadjaheitmann: @ares Could you help us out here? It is about the implementation of the Puppet Metrics table https://user-images.githubusercontent.com/51960547/278665889-e07d582f-fa86-4861-ad21-0f6a39650e2c.png

MariaAga: From the screenshot, we can also use `-` instead of n/a

nadjaheitmann: It is actually `--`, the same as on the Reports tab. That is what I currently have implemented.

MariaAga: I mean something like ` <TableText>{value || --}</TableText>`

nadjaheitmann: What is the difference between `<TableText>{value || --}</TableText>` and setting the default value in function parameter?

MariaAga: None, I missed the default in the function sorry

nadjaheitmann: Is there anything else I should fix?

webpack/src/Extends/Host/PuppetTab/SubTabs/Reports/components/ConfigStatusCard.js
```diff
@@ -0,0 +1,177 @@
...
+          metrics.corrective_change
+        )}
+      </Tr>
+      <Tr>
+        {createPuppetMetricsTableElement(__('Skipped'), metrics.skipped)}
+        {createPuppetMetricsTableElement(
+          __('Out of sync'),
+          metrics.out_of_sync
+        )}
+        <b>{createPuppetMetricsTableElement(__('Total'), metrics.total)}</b>
```
MariaAga, line 120: I get this warning: ` <b> cannot appear as a child of <tr>.` 
Also it causes the `Total` to be misaligned with the rest

nadjaheitmann: Right, this should not be there :see_no_evil: 

Shall I implement the `Total` in bold font? I guess this would need some css or is there an other way?

MariaAga: just some css using `font-weight`

nadjaheitmann: Fixed. Also refactored the code.

nadjaheitmann: @MariaAga Any more things I should fix?

MariaAga: No, all looks good to me

nadjaheitmann: @stejskalleos I think, we can move forward and merge, then.

stejskalleos:
Thanks @nadjaheitmann @MariaAga !

PR #380, opened 2024-01-09
Fix rubocop rules

archanaserver:


foreman_puppet.gemspec
```diff
@@ -13,4 +13,6 @@ Gem::Specification.new do |s|
 
   s.files = Dir['{app,config,db,lib,locale,webpack}/**/*'] + ['LICENSE', 'Rakefile', 'README.md', 'package.json']
   s.test_files = Dir['test/**/*']
+
+  s.required_ruby_version = '>= 2.7', '< 4'
```
stejskalleos, line 17: Why `< 4`? It should be just `~> 2.7.0`

.rubocop.yml
```diff
@@ -8,6 +8,12 @@ inherit_mode:
   merge:
     - Exclude
 
+AllCops:
+  TargetRubyVersion: '2.7'
+  Exclude:
+    - 'test/**/*'
```
ekohl, line 14: I don't think it's good to exclude RuboCop on all tests.

foreman_puppet.gemspec
ekohl: `~> 2.7.0` would prevent it from running on Ruby 3, which is something we'd like to do.

ekohl (approved):
Lgtm. You could argue we want to first test Ruby 3, but without the line it'd also install on Ruby 3. I'd go ahead with this but leaving to @stejskalleos for a final approval

stejskalleos (approved):
:green_apple: **LGTM**

PR #381, opened 2024-01-11
 Rewrite GHA workflow using reusable actions

archanaserver:
Updated GitHub Action workflow using shared Rubocop and Foreman plugin test actions.

https://github.com/theforeman/actions#rubocop
https://github.com/theforeman/actions#foreman-plugin-tests

### DIFF
--- a/.github/workflows/ruby_tests.yml
+++ b/.github/workflows/ruby_tests.yml
@@ -1,97 +1,30 @@
 name: Ruby Testing
+
 on:
   pull_request:
   push:
     branches:
-    - master
-env:
-  RAILS_ENV: test
-  DATABASE_URL: postgresql://postgres:@localhost/test
-  DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: true
+      - 'master'
+
+concurrency:
+  group: ${{ github.ref_name }}-${{ github.workflow }}
+  cancel-in-progress: true
+
 jobs:
   rubocop:
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
+    with:
+      command: bundle exec rubocop -P --format github
+      
+  test:
+    name: Ruby
+    needs: rubocop
     runs-on: ubuntu-latest
+
     env:
       BUNDLE_FOREMAN: '0'
-      BUNDLE_WITHOUT: development:test
+
     steps:
-      - uses: actions/checkout@v2
-      - name: Setup Ruby
-        uses: ruby/setup-ruby@v1
-        with:
-          ruby-version: 2.7
-          bundler-cache: true
-      - name: Run rubocop
-        run: bundle exec rubocop -P --format github
-  test_ruby:
-    runs-on: ubuntu-latest
-    needs: rubocop
-    env:
-      BUNDLE_WITHOUT: journald:development:console:libvirt
-    services:
-      postgres:
-        image: postgres:12.1
-        ports: ['5432:5432']
-        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
-    strategy:
-      fail-fast: true
-      matrix:
-        foreman-repo: [theforeman/foreman]
-        foreman-core-branch: [develop]
-        ruby-version: [2.7]
-        node-version: [12]
-    steps:
-      - run: |
-          sudo apt-get update
-          sudo apt-get install build-essential libcurl4-openssl-dev zlib1g-dev libpq-dev
-      - name: Install chromedriver
-        uses: nanasess/setup-chromedriver@master
-      - uses: actions/checkout@v2
-        with:
-          repository: ${{ matrix.foreman-repo }}
-          ref: ${{ matrix.foreman-core-branch }}
-      - uses: actions/checkout@v2
-        with:
-          path: foreman_puppet
-      - name: Setup Bundler
-        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ruby_tests.yml
```diff
@@ -4,94 +4,26 @@ on:
...
-          npm install
-          bundle exec rake webpack:compile
-      - name: Run plugin tests
-        if: github.event_name != 'push'
-        run: |
-          bundle exec rake test:foreman_puppet
-      - name: Run access permission test
-        if: ${{ github.event_name != 'push' }}
-        run: |
-          bundle exec rake test TEST="test/unit/foreman/access_permissions_test.rb"
```
archanaserver, line 91: This steps has not yet executed in the reusable action.

.github/workflows/ruby_tests.yml
```diff
@@ -4,94 +4,26 @@ on:
...
-        foreman-repo: [theforeman/foreman]
-        foreman-core-branch: [develop]
-        ruby-version: [2.7]
-        node-version: [12]
-    steps:
-      - run: |
-          sudo apt-get update
-          sudo apt-get install build-essential libcurl4-openssl-dev zlib1g-dev libpq-dev
-      - name: Install chromedriver
-        uses: nanasess/setup-chromedriver@master
```
archanaserver, line 48: @ekohl i can't see this step execution in the reusable action, do we want to execute this one? i'm not aware of this one.

.github/workflows/ruby_tests.yml
ekohl: Related discussion: https://github.com/theforeman/actions/pull/16

.github/workflows/ruby_tests.yml
ekohl: We need a browser for the acceptance tests. I think a separate container that runs selenium together with https://github.com/theforeman/foreman/pull/9952 could be an alternative. It would be best to track this by opening an issue on https://github.com/theforeman/actions or at least mention it in https://github.com/theforeman/actions/issues/1

.github/workflows/ruby_tests.yml
```diff
@@ -4,94 +4,26 @@ on:
...
 env:
   RAILS_ENV: test
   DATABASE_URL: postgresql://postgres:@localhost/test
   DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: true
+
 jobs:
   rubocop:
-    runs-on: ubuntu-latest
-    env:
-      BUNDLE_FOREMAN: '0'
```
ekohl, line 15: This one needs to remain for the RuboCop action. At least with the current Gemfile. It's a very complex Gemfile and it's questionable if it's a good idea. I'd prefer if it was much closer to our template (https://github.com/theforeman/foreman_plugin_template/blob/master/Gemfile) and only listed what it actually needed (like RuboCop).

archanaserver: @ekohl i didn't clearly get it, could you please elaborate further?

ekohl: If you look at https://github.com/theforeman/foreman_puppet/blob/master/Gemfile you can see it has different behavior depending on whether the environment variable `BUNDLE_FOREMAN` has a specific value. Since you're leaving this out, it flips to the other behavior, which causes the failures you see.

I was then asking myself if this behavior actually needs to remain, but to move ahead faster I'd start with setting the environment variable as it set in the current CI.

nadjaheitmann: Sounds good to me. Might be a good idea to change it soon, otherwise people will stumble across it again, eventually.

.github/workflows/ruby_tests.yml
```diff
@@ -4,94 +4,27 @@ on:
   push:
     branches:
     - master
-env:
-  RAILS_ENV: test
-  DATABASE_URL: postgresql://postgres:@localhost/test
-  DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: true
```
archanaserver, line 10: Dropped this change, we already have this in reusable `workflows/foreman_plugin.yml`

.github/workflows/ruby_tests.yml
archanaserver: https://github.com/theforeman/actions/issues/38

archanaserver:
@ekohl how the workflow is not getting triggered? i wonder if any checks i'm missing here :/

Gemfile
```diff
@@ -1,22 +1,7 @@
 if ENV.fetch('BUNDLE_FOREMAN', '1') == '0'
```
ekohl, line 1: I thought you'd drop this too.

Gemfile
```diff
@@ -1,22 +1,7 @@
...
   source 'https://rubygems.org'
-else
-  foreman_path = Dir['./foreman', '../foreman', '../../foreman']
-  foreman_path.map! { |path| File.join(path, 'Gemfile') }
-  foreman_gemfile = foreman_path.detect { |p| File.exist?(p) }
-  raise 'Foreman have not been found!' unless foreman_gemfile
 
-  foreman_gemfile = File.expand_path(foreman_gemfile)
-  eval_gemfile foreman_gemfile
+  gemspec name: 'foreman_puppet'
```
ekohl, line 4: Usually it can detect this automatically
```suggestion
  gemspec
```

ekohl (changes requested):

ekohl (approved):

PR #383, opened 2024-01-29
Fix tests for Ruby 3.0

ofedoren:
~Please, don't merge just yet. Foreman on Ruby 3.0 is still in progress.~

Depends on https://github.com/theforeman/foreman_puppet/pull/381 ~. Included here.~


ofedoren:
It seems there were some issues with GHA definition for this repo. I changed it to be in line with the others definitions proved to be working, so we have some runs here.

PR #389, opened 2024-03-04
Release gem GH action

stejskalleos:


nadjaheitmann:
I like the topic that much that I opened another PR for this :see_no_evil:  https://github.com/theforeman/foreman_puppet/pull/395 

I think, we should merge one of them :)

.github/workflows/release_gem.yml
```diff
@@ -0,0 +1,14 @@
+name: Release
+
+on:
+  create:
+    ref_type: tag
```
evgeni, line 5:
```suggestion
  push:
    tags:
      - 'v*'
```

It's been said this is better (see https://github.com/theforeman/actions/?tab=readme-ov-file#gem-release)

.github/workflows/release_gem.yml
```diff
@@ -0,0 +1,14 @@
...
+
+jobs:
+  release:
+    name: Release gem
+    uses: theforeman/actions/.github/workflows/release-gem.yml@v0
+    with:
+      allowed_owner: theforeman
+    secrets:
+      api_key: ${{ secrets.RUBYGEM_API_KEY }}
```
evgeni, line 14: created

evgeni (changes requested):

.github/workflows/release_gem.yml
```diff
@@ -0,0 +1,14 @@
+name: Release
```
nadjaheitmann, line 1: I wonder why the github actions are never prepended with `---`  ?

evgeni: people being lazy? :)

nadjaheitmann: I propose to add another action as done here, then... ;) https://github.com/theforeman/smart_proxy_salt/blob/master/.github/workflows/yamllint.yaml

stejskalleos: > people being lazy? :)

Yes.

evgeni (approved):

Issue #382, opened 2024-01-25
Adapt foreman_puppet tests to run with remote_webdriver


dosas:
When executing foreman_puppet integration tests with the [remote webdriver](https://github.com/theforeman/foreman/pull/9952)

I get the following error:

```
test_0005_edit page(Minitest::Result) [/builds/Engineering/orcharhino/dependencies/foreman_puppet/foreman/vendor/ruby/ruby/2.7.0/gems/selenium-webdriver-4.9.0/lib/selenium/webdriver/remote/bridge.rb:411]:
Selenium::WebDriver::Error::WebDriverError: You are trying to work with something that isn't a file.
foreman_puppet/test/integration/foreman_puppet/smartclass_parameter_js_test.rb:77:in `block in <class:SmartclassParameterJSTest>'

```

In this [file](https://github.com/theforeman/foreman_puppet/blob/master/test/integration/foreman_puppet/smartclass_parameter_js_test.rb#L77)
```
fill_in 'puppetclass_lookup_key_description', with: 'test'
```
Is looking for a file called test

AFAIK there are to possibilities to fix this:

1. Disable the file detector for this test
```
page.driver.browser.file_detector = nil if page.driver.browser.respond_to?(:file_detector=)
```

2. Change fill_in to something that isn't a file
```
fill_in 'puppetclass_lookup_key_description', with: 'asdf'
```

Let me know which one you prefer then I can open a PR

nadjaheitmann:
@stejskalleos Any idea what is a good solution for this?

dosas:
https://github.com/theforeman/foreman_puppet/pull/392

PR #392, opened 2024-03-13
Fixes #37251: Error occuring for remote browser

dosas:
Selenium::WebDriver::Error::WebDriverError: You are trying to work with something that isn't a file.

### DIFF
--- a/test/integration/foreman_puppet/smartclass_parameter_js_test.rb
+++ b/test/integration/foreman_puppet/smartclass_parameter_js_test.rb
@@ -74,6 +74,8 @@ module ForemanPuppet
       within(:xpath, '//table') do
         click_link 'ssl'
       end
+      # see: https://projects.theforeman.org/issues/37251
+      page.driver.browser.file_detector = nil if page.driver.browser.respond_to?(:file_detector=)
       fill_in 'puppetclass_lookup_key_description', with: 'test'
       fill_in 'puppetclass_lookup_key_default_value', with: 'false'
       assert_submit_button(puppetclass_lookup_keys_path)


test/integration/foreman_puppet/smartclass_parameter_js_test.rb
```diff
@@ -74,6 +74,7 @@ class SmartclassParameterJSTest < IntegrationTestWithJavascript
       within(:xpath, '//table') do
         click_link 'ssl'
       end
+      page.driver.browser.file_detector = nil if page.driver.browser.respond_to?(:file_detector=)
```
stejskalleos, line 77:
```suggestion
      page.driver.browser&.file_detector = nil
```

stejskalleos, line 77: If this is a workaround for buggy behavior, I would suggest adding a comment with a link to the Redmine or a bug report.
Also, shouldn't this be a part of before/after test setup actions?

stejskalleos (changes requested):

test/integration/foreman_puppet/smartclass_parameter_js_test.rb
dosas: AFAIK this is a problem with the selenium ruby bindings https://stackoverflow.com/questions/70441796/selenium-webdriver-for-aws-device-farm-error-when-sending-period-keystroke-t

another more implicit fix is described here https://github.com/theforeman/foreman_puppet/issues/382 to just rename `test` to something that is not locally available

test/integration/foreman_puppet/smartclass_parameter_js_test.rb
```diff
@@ -74,6 +74,7 @@ class SmartclassParameterJSTest < IntegrationTestWithJavascript
       within(:xpath, '//table') do
         click_link 'ssl'
       end
+      page.driver.browser&.file_detector = nil
```
stejskalleos, line 77: Can you add a comment  to the code as well, with comment link to the Redmine? So it's immediately clear what's the reasoning behind this workaround.

dosas: done

test/integration/foreman_puppet/smartclass_parameter_js_test.rb
```diff
@@ -74,6 +74,8 @@ class SmartclassParameterJSTest < IntegrationTestWithJavascript
       within(:xpath, '//table') do
         click_link 'ssl'
       end
+      # see: https://projects.theforeman.org/issues/37251
+      page.driver.browser&.file_detector = nil
```
stejskalleos, line 78: Hmm, tests are failing now on:
```
NoMethodError: undefined method `file_detector=' for #<Selenium::WebDriver::Chrome::Driver:0x..fda95f2e33cc25c50 browser=:chrome_headless_shell>
    foreman_puppet/test/integration/foreman_puppet/smartclass_parameter_js_test.rb:78:in `block in <class:SmartclassParameterJSTest>'
rake aborted!
```

But why? There is the safe guard ...

dosas: I don't know. I would propose to switch back to the version that worked.

stejskalleos: Yop, agree ... Sorry for the not working suggestion

dosas: Nevermind. I thought it would work too. I updated the PR.

stejskalleos:
Thanks @dosas

PR #393, opened 2024-03-26
Fixes #37299 - Fix puppet class assignment

adamruzicka:
In case the puppet facet does not exist


stejskalleos:
@adamruzicka I see it's in the draft state, is it ready for review?

stejskalleos (approved):
:green_apple: **LGTM**

PR #387, opened 2024-02-20
Fixes #386 - Change search parameter to puppetclass

nadjaheitmann:

Issue #386, opened 2024-02-18
Puppet Classes page has incorrect link to host search page to view hosts with class


bigwillystyle42:
When you're on the Puppet Classes page: www.example.com/foreman_puppet/puppetclasses the hosts column shows the number of hosts with a class. Clicking the link in Foreman 3.9.1 takes you to the hosts page, displays all hosts, and displays an error message "Field 'class' not recognized for searching!"

Those links look like this:

https://www.example.com/hosts?search=class+%3D+CLASSNAMEHERE

But I think they should be puppetclass not class

https://www.example.com/hosts?search=puppetclass+%3D+CLASSNAMEHERE

PR #394, opened 2024-04-01
Fix querying for hosts using puppet class

bmagistro:
When trying to query foreman for hosts that are using a particular puppet class, the query does not return the desired results.  In the process of tracking this down we also encountered #386 / #387 .

----

Given the puppet class, it correctly lists that there are two hosts using that class.
Screenshot 2024-03-30 at 22 32 08

If we follow the link we are presented with the following hosts using the class
Screenshot 2024-03-30 at 22 32 38

This was tracked down to a missing sql join.  We got there by raising errors and doing some explains.  The queries below were ran directly to help us debug.

```
foreman=# SELECT "config_groups"."id" FROM "config_groups" INNER JOIN "config_group_classes" ON "config_group_classes"."config_group_id" = "config_groups"."id" INNER JOIN "puppetclasses" ON "puppetclasses"."id" = "config_group_classes"."puppetclass_id" WHERE (puppetclasses.name = 'profiles::lanes::protected') ORDER BY config_groups.name;
 id 
----
 10
(1 row)


foreman=# SELECT "host_config_groups"."host_id" FROM "host_config_groups" WHERE "host_config_groups"."host_type" = 'ForemanPuppet::HostPuppetFacet' AND "host_config_groups"."config_group_id" = 10;
 host_id 
---------
     270
     183
(2 rows)
```
The above host_ids are actually host facet ids so the it tries to query these as host ids resulting in the mismatch.
Adding the below sql before running the query for hosts we get to what we expect
```
foreman=# select * from host_puppet_facets where id = 270 or id = 183;
 id  | host_id | environment_id | puppet_proxy_id | created_at |         updated_at         
-----+---------+----------------+-----------------+------------+----------------------------
 270 |     430 |              1 |                 |            | 
 183 |     460 |              1 |                 |            | 2022-07-03 18:29:34.534469
(2 rows)

foreman=# select id, name from hosts where id = 430 or id = 460;
 id  |                    name                    
-----+--------------------------------------------
 460 | trust-0002.dev.***
 430 | trust-0001.dev.***
(2 rows)
```

Using the changed code that is in the PR, and performing the same query we get the desired results
Screenshot 2024-03-31 at 21 40 05

### DIFF
--- a/app/models/concerns/foreman_puppet/extensions/host.rb
+++ b/app/models/concerns/foreman_puppet/extensions/host.rb
@@ -24,10 +24,9 @@ module ForemanPuppet
           conditions = sanitize_sql_for_conditions(["puppetclasses.name #{operator} ?", value_to_sql(operator, value)])
           config_group_ids = ForemanPuppet::ConfigGroup.joins(:puppetclasses).where(conditions).pluck(:id)
           host_ids         = ::Host.authorized(:view_hosts).joins(puppet: :puppetclasses).where(conditions).distinct.pluck(:id)
-          host_ids        += ForemanPuppet::HostConfigGroup
-                             .where(host_type: 'ForemanPuppet::HostPuppetFacet')
-                             .where(config_group_id: config_group_ids)
-                             .pluck(:host_id)
+          host_ids        += ForemanPuppet::HostPuppetFacet.joins(:host_config_groups)
+                                                           .where(host_config_groups: { config_group_id: config_group_ids })
+                                                           .pluck(:host_id)
           hostgroup_ids = ::Hostgroup.unscoped.with_taxonomy_scope.joins(puppet: :puppetclasses).where(conditions).map(&:subtree_ids)
           if config_group_ids.any?
             hostgroup_cg_ids = ForemanPuppet::HostgroupPuppetFacet.joins(:host_config_groups)


stejskalleos:
@bmagistro Thanks for the contribution, can you please fix the rubocop findings?

app/models/concerns/foreman_puppet/extensions/host.rb
```diff
@@ -24,10 +24,10 @@ def search_by_puppetclass(_key, operator, value)
           conditions = sanitize_sql_for_conditions(["puppetclasses.name #{operator} ?", value_to_sql(operator, value)])
           config_group_ids = ForemanPuppet::ConfigGroup.joins(:puppetclasses).where(conditions).pluck(:id)
           host_ids         = ::Host.authorized(:view_hosts).joins(puppet: :puppetclasses).where(conditions).distinct.pluck(:id)
-          host_ids        += ForemanPuppet::HostConfigGroup
-                             .where(host_type: 'ForemanPuppet::HostPuppetFacet')
-                             .where(config_group_id: config_group_ids)
-                             .pluck(:host_id)
+          host_ids        += ForemanPuppet::HostPuppetFacet.joins(:host_config_groups)
+                                                           .where(host_config_groups: { host_type: 'ForemanPuppet::HostPuppetFacet' })
```
ekohl, line 28: Isn't this redundant now that you made it a proper join?
```suggestion
```

bmagistro: At the time we were just trying to get it back to working and didn't check for/think about redundant items in the sql.  Adding this to our locally patched version and testing, I would agree it can be removed.

nadjaheitmann (changes requested):
@bmagistro Can you please squash your commits?

nadjaheitmann (approved):

PR #399, opened 2024-05-20
Add customizable column to React HostsIndex page

jeremylenz:
Add a "Puppet env" column to the new React All Hosts page.

For a demo with some more context, see https://www.youtube.com/watch?v=bzZEOiJOrCg&t=1808s

This will bring "column parity" for foreman_puppet between the legacy and new All Hosts page.

### DIFF
--- a/webpack/global_index.js
+++ b/webpack/global_index.js
@@ -1,4 +1,6 @@
 import { registerReducer } from 'foremanReact/common/MountingService';
+import { registerColumns } from 'foremanReact/components/HostsIndex/Columns/core';
+import { translate as __ } from 'foremanReact/common/I18n';
 import reducers from './src/reducers';
 import { registerFills } from './src/Extends/Fills';
 import { registerLegacy } from './legacy';
@@ -9,3 +11,22 @@ registerReducer('puppet', reducers);
 registerFills();
 // TODO: the checkForUnavailablePuppetclasses is very nasty
 registerLegacy();
+
+// register columns for React hosts index page
+const puppetHostsIndexColumns = [
+  {
+    columnName: 'environment',
+    title: __('Puppet env'),
+    isSorted: true,
+    wrapper: hostDetails => hostDetails.environment_name,
+    weight: 2700,
+  },
+];
+
+puppetHostsIndexColumns.forEach(column => {
+  column.tableName = 'hosts';
+  column.categoryName = 'Puppet';
+  column.categoryKey = 'puppet';
+});
+
+registerColumns(puppetHostsIndexColumns);


chris1984 (approved):
I don't have merge access, but code looks fine.

webpack/global_index.js
```diff
@@ -1,4 +1,6 @@
 import { registerReducer } from 'foremanReact/common/MountingService';
+import { registerColumns } from 'foremanReact/components/HostsIndex/Columns/core';
+import { translate as __ } from 'foremanReact/common/i18n';
```
nadjaheitmann, line 3:
```suggestion
import { translate as __ } from 'foremanReact/common/I18n';
```

webpack/global_index.js
```diff
@@ -9,3 +11,16 @@ registerReducer('puppet', reducers);
 registerFills();
 // TODO: the checkForUnavailablePuppetclasses is very nasty
 registerLegacy();
+
+// register columns for React hosts index page
+const puppetHostsIndexColumns = [
+  {
+    columnName: 'environment',
```
nadjaheitmann, line 18:
```suggestion
    columnName: 'puppet_environment',
```

Just to make sure it does not get confused with lifecycle_environment in Katello.

nadjaheitmann:
> @stejskalleos how does this look?

@stejskalleos I got a new dev setup working after all, so I could do the review!

webpack/global_index.js
jeremylenz: Would be nice, but

1) The existing column name for the legacy page is already `environment` and I need to keep it the same, otherwise users' table preferences won't carry over to the new page.
2) the Katello column is `lifecycle_environment` so in this case we won't have a conflict.

webpack/global_index.js
```diff
@@ -9,3 +11,16 @@ registerReducer('puppet', reducers);
...
+    title: __('Puppet env'),
+    isSorted: true,
+    wrapper: hostDetails => hostDetails.environment_name,
+    weight: 2700,
+  },
+];
+
+registerColumns(puppetHostsIndexColumns);
```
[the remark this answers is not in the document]
jeremylenz: You'd think being the one who wrote that originally, I'd remember it :grinning:

jeremylenz:
fixed lint

webpack/global_index.js
nadjaheitmann: I see. I guess that is a leftover from the time that puppet was still part of core.

nadjaheitmann (approved):

nadjaheitmann:
Which is the minimum Foreman version that supports this fix?

jeremylenz:
I believe it should be 3.11: https://projects.theforeman.org/issues/37293

Issue #397, opened 2024-05-07
"change Puppet Master" option changes both Puppet Capsule and Puppet CA Capsule at same time


stejskalleos:
Description of problem:
Reporting this in response to findings from https://bugzilla.redhat.com/show_bug.cgi?id=2155083#c21 testing and discussion

How reproducible:
Always

Steps to Reproduce:
1. Enabled Puppet on Satellite and register a few hosts to Satellite.
2. Navigate to Hosts -> All Hosts -> Select Host -> Select Action -> Change Puppet Master
3.

Actual results:
"change Puppet Master" option changes both Puppet Capsule and Puppet CA Capsule at same time.
and notifies that only puppet capsule changed always like
"The Puppet Capsule of the selected hosts was set to sat.example.com

Expected results:
Either this option shouldn't change both as we've "Change Puppet CA" option already
OR we should notify correctly that both fields changed.

Additional info:
1. When we've both Puppet Capsule and Puppet CA Capsule fields set already on host, and if we try "change Puppet Master" option then it only changes Puppet Capsule field for host, not Puppet CA Capsule as expected
2. "Change Puppet CA" works correctly and it updates on Puppet CA Capsule field on host

https://issues.redhat.com/browse/SAT-23107

Issue #396, opened 2024-05-07
"Clear Puppet/PuppetCA Capsule" option doesn't work on host with Change Puppet Master or Change Puppet CA bulk action


stejskalleos:
Description of problem:
Reporting this in response to findings from https://bugzilla.redhat.com/show_bug.cgi?id=2155083#c21 testing and discussion

Steps to Reproduce:
1. Enabled Puppet on Satellite and register a few hosts to Satellite.
2. Navigate to Hosts -> All Hosts -> Select Host -> Select Action -> Change Puppet Master or Change Puppet CA
3. Try *Clear Puppet Capsule* or *Clear Puppet CA Capsule* option from the dropdown to clear assigned 

Actual results:
Submit button is disabled

Expected results:
Submit button enabled, and it clears the field on the host for Puppet Capsule or Puppet CA Capsule

https://issues.redhat.com/browse/SAT-23106

nadjaheitmann:
As the All hosts page is currently under construction: 
@stejskalleos  Do you think it is worth investigating this for the old 'All hosts' page? What is the time frame when the new page will be available?

nadjaheitmann:
@stejskalleos Pinging you again :) see above

stejskalleos:
If possible, I would fix the issue on both; if not, then I would prefer the new page.

PR #401, opened 2024-06-28
Fixes #36207 - assign puppet attributes from hostgroup

sbernhard:


app/models/concerns/foreman_puppet/extensions/host.rb
```diff
@@ -55,6 +55,15 @@ def search_by_config_group(_key, operator, value)
       end
 
       module PrependedMethods
+        def environment_id=(id)
+          self.environment = ForemanPuppet::Environment.find(id)
+          save
+        end
+
+        def inherited_attributes
```
sbernhard, line 63: This method is missing for sure so that attributes from hostgroup are used. Same as in katello, openscap, salt...

app/models/concerns/foreman_puppet/extensions/host.rb
```diff
@@ -55,6 +55,15 @@ def search_by_config_group(_key, operator, value)
       end
 
       module PrependedMethods
+        def environment_id=(id)
+          self.environment = ForemanPuppet::Environment.find(id)
+          save
```
sbernhard, line 60: Is this good or does it break something to have a ènvironment_id` method? 
can the `save` be improved?

app/models/concerns/foreman_puppet/extensions/hostgroup.rb
```diff
@@ -30,6 +30,26 @@ class << self
           ext_method: :search_by_config_group
       end
 
+      def inherited_puppet_hostgroup_attribute(attribute)
+        if ancestry.present?
```
sbernhard, line 34: grabed and adapted from foreman_openscap

nadjaheitmann:
Looks like it breaks the tests. Can you please verify? 

Also could this be related to #396 and #397?

app/models/concerns/foreman_puppet/extensions/host.rb
sbernhard: Do you know @ekohl (or know somehow you maybe can answer this question), why this method did not exist in puppet but is used for Katello, OpenSCAP, Salt. 
How does this work, if you create a host which is using a hostgroup, that this host is using the "templated values" from hostgroup without this code?

nadjaheitmann: As this is related to facets (environment_id is provided by the puppet facet), maybe we can also ping @ShimShtein

PR #403, opened 2024-07-16
Fixes #37652 - change jquery size to length

MariaAga:
Tested that the warning still appears:
Have 2 puppet envs
Create a host with env1 and class1 from it
Edit host to have env2, the warning appears 

.size is deprecated and length can be used instead

### DIFF
--- a/webpack/src/foreman_puppet_host_form.js
+++ b/webpack/src/foreman_puppet_host_form.js
@@ -110,7 +110,7 @@ export function checkForUnavailablePuppetclasses() {
       ${warningMessage}
     </span>`;
 
-  if (unavailableClasses.size() > 0) {
+  if (unavailableClasses.length > 0) {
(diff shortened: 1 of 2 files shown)


nadjaheitmann:
Thanks  @MariaAga for the fix. There are also calls to `size` in the tests https://github.com/theforeman/foreman_puppet/blob/master/webpack/src/foreman_puppet_host_form.test.js . Should they be fixed as well?

MariaAga:
Good catch thanks! changed the test as well

nadjaheitmann:
> Good catch thanks! changed the test as well

Now the tests are complaining... I would ignore the unrelated one, but the warnings seem valid.

MariaAga:
fixed the test lint

nadjaheitmann (approved):
Tested and works fine. Thanks @MariaAga

PR #404, opened 2024-07-19
switch parseJSON to JSON.parse

MariaAga:
$.parseJSON will be deprecated, and JSON.parse does the same thing, but as part of JS and not jquery


nadjaheitmann (approved):
LGTM. Tested and everything seems to work just fine.

PR #407, opened 2024-08-19
i18n - pulling from tx

stejskalleos:


locale/fr/foreman_puppet.po
```diff
@@ -41,12 +43,12 @@ msgstr "%%{model} aynat pour id '%%{id}' n'a pas été trouvé"
 
 msgid "%{name} has %{num_tag} class"
 msgid_plural "%{name} has %{num_tag} classes"
-msgstr[0] "%%{name} a %{num_tag} classe"
-msgstr[1] "%%{name} a %{num_tag} classes"
-msgstr[2] "%%{name} a %{num_tag} classes"
+msgstr[0] "%%{name} a %%{num_tag} classe"
+msgstr[1] "%%{name} a %%{num_tag} classes"
+msgstr[2] "%%{name} a %%{num_tag} classes"
```
ekohl, line 48: This doesn't look good. I've fixed the various places in translations where `%%` was used. Please pull translations again.

locale/ja/foreman_puppet.po
```diff
@@ -37,7 +36,7 @@ msgstr "id '%{id}' の %{model} が見つかりませんでした"
 
 msgid "%{name} has %{num_tag} class"
 msgid_plural "%{name} has %{num_tag} classes"
-msgstr[0] "%{name} には ％{num_tag}クラスが含まれます"
+msgstr[0] "%%{name} には ％%{num_tag} クラスが含まれます"
```
ekohl, line 39: This also doesn't look good so I fixed them manually.

locale/pt_BR/foreman_puppet.po
```diff
@@ -251,10 +253,10 @@ msgid "ENC environment"
 msgstr "Ambiente ENC"
 
 msgid "Edit %s"
-msgstr "Editar %s"
+msgstr "Editar"
```
ekohl, line 256: This isn't right so I fixed them up.

locale/zh_CN/foreman_puppet.po
```diff
@@ -38,7 +39,7 @@ msgstr "找不到 ID 为 '%{id}' 的 %{model}"
 
 msgid "%{name} has %{num_tag} class"
 msgid_plural "%{name} has %{num_tag} classes"
-msgstr[0] "%{name} 具有 ％{num_tag}类"
+msgstr[0] "%%{name} 具有 ％%{num_tag} 类"
```
ekohl, line 42: Same as others: I fixed these up.

ekohl (changes requested):

locale/fr/foreman_puppet.po
stejskalleos: I pulled the translations but nothing changed.

ekohl: I can reproduce that it's not pulling correctly, but at this point it's unclear why. When I run `tx pull` manually I see it creates an updated `locale/fr/foreman_puppet.edit.po` (so `make -C locale tx-pull` is correct) but then something later I see:
```
rm -f /home/ekohl/dev/foreman_puppet/locale/fr/foreman_puppet.edit.po
cp /home/ekohl/dev/foreman_puppet/locale/fr/foreman_puppet.po /home/ekohl/dev/foreman_puppet/locale/fr/foreman_puppet.edit.po
touch /home/ekohl/dev/foreman_puppet/locale/fr/foreman_puppet.po.time_stamp
```

That throws away all the updated translations. That gets executed as part of `rake plugin:gettext[$plugin]`, which is a thin wrapper around https://github.com/grosser/gettext_i18n_rails.

ekohl: I think I found the fix: https://github.com/theforeman/foreman_plugin_template/pull/65.

Now that I see it in action, I see it also clears out the Swedish translations, which is probably because we set the minimum threshold to pull to 50%. That's an unintended consequence of https://github.com/theforeman/foreman_puppet/commit/9139cb596508bd1f7fccfd55c188600ac37c89c8 that was previously masked by the update bug.

stejskalleos:
Will continue in https://github.com/theforeman/foreman_puppet/pull/409/files

PR #410, opened 2024-09-20
Fixes #37851 - Support Rails 7.0

ofedoren:


ekohl (commented):
These changes look like they're 6.1 compatible and I'd like to merge them.

ofedoren:
Dropped the last commit.

ekohl (approved):

nadjaheitmann:
I read through the [community post](https://community.theforeman.org/t/path-to-ruby-3-0-3-1-el9-and-ubuntu-22-04/30295/41) and I am not sure that I understand until when we need Rails 7 support. Can you clarify, please?

ekohl:
@nadjaheitmann this addresses some deprecations that already makes it compatible with Rails 7 while we can also ship with Rails 6.1, meaning we don't have to coordinate releases as much.

You should read https://community.theforeman.org/t/rails-7-upgrade/37771 though. We split off the effort, but the important thing to note is that Rails 6.1 is already out of security support. It's urgent that we upgrade.

PR #411, opened 2024-11-05
Fixes #37979 - Fix puppet classes import filter when using Ruby 3+ and Regexp/Symbol in YAML file

neomilium:
See details here: https://projects.theforeman.org/issues/37979


nadjaheitmann:
Is this compatible with older Ruby versions?

neomilium:
> Is this compatible with older Ruby versions?

I just amend my commit, this should now be compatible with older Ruby versions.

neomilium:
@nadjaheitmann Up?

neomilium:
@stejskalleos Done. Thanks.

ekohl (commented):
A difference is that they changed the `load_file` method to really be the `safe_load_file` method. Sadly on Psych bundled in Ruby 2.7, `safe_load_file` doesn't exist.

A way to avoid a specific version check is:
```ruby
if File.exist?(ignored_file_path)
  # Ruby 3+
  if YAML.respond_to(:safe_load_file)
    YAML.safe_load_file(ignored_file_path, permitted_classes: [Symbol, Regexp])
  else
    YAML.load_file(ignored_file_path)
  end
else
  {}
end
```

Once we drop Ruby 2.7 support, you can drop the `load_file` call.

neomilium:
@ekohl Your suggestion is now applied. Thanks.

ekohl (approved):
I'll admit I haven't tested this myself, but I'm approving CI to run.

nadjaheitmann:
@stejskalleos Ewoud approved, I suggest to merge this.

PR #412, opened 2024-11-06
Update compatibility matrix

nadjaheitmann:


README.md
```diff
@@ -33,7 +33,8 @@ Some features will remain in core:
 
 |Foreman version|Plugin version|Notes                                     |
 |---------------|--------------|------------------------------------------|
-| >= 3.11       | ~> 7.0       | Required                                 |
+| >= 3.13       | ~> 8.0       | Required                                 |
```
stejskalleos, line 36: According to our [nightly docs](https://docs.theforeman.org/nightly/Managing_Configurations_Puppet/index-katello.html#supported-puppet-versions-and-system-requirements_managing-configurations-puppet), the 7.0 version is still supported

nadjaheitmann: Then we should update the docs, because foreman_puppet 8.0 is the one that adds Zeitwerk.

nadjaheitmann: Oh, I misunderstood your comment. The docs is about the Puppet version but this is not related to how foreman and foreman_puppet plugin match. Which this PR is about.

stejskalleos: Oh I see it now, sorry! Let's get this merged then.

PR #413, opened 2024-11-19
Fix environment param in smart proxies controller

mtancoigne:
Hello!

We noticed an issue in the way environments parameter is used in the proxies controller:
routes `/api/environments/:environment/` handles environment name as `:environment` param, while `/api/environments/:environment/smart_proxies/:smart_proxy_id/import_puppetclasses` searches for a numerical identifier.

This leads to _all_ the environments being imported when a name is specified.

To stay consistent, the fix uses the same method as the one used on `/api/environments/:environment/` to find the environment.

### DIFF
--- a/app/controllers/concerns/foreman_puppet/extensions/api_smart_proxies_controller.rb
+++ b/app/controllers/concerns/foreman_puppet/extensions/api_smart_proxies_controller.rb
@@ -140,7 +140,7 @@ module ForemanPuppet
       end
 
       def find_optional_environment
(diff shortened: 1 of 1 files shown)


neomilium:
@ekohl Could you unlock the CI run? plz

nadjaheitmann:
Looks good to me. What do you think @stejskalleos?

stejskalleos:
Thanks @mtancoigne @nadjaheitmann

PR #414, opened 2024-12-03
Zeitwerk & ProvisioningTemplate extensions

stejskalleos:
Removing Puppet's associations from snippets
stopped working with the latest changes
in Zeitwerk and the way we load the classes.


nadjaheitmann (approved):
Looks good, thank you @stejskalleos ! 
I think we can also remove EL8 build but we can do it in a follow-up PR. 

Edit: Just saw the other PR :)

PR #415, opened 2024-12-03
Remove packit for EL8

stejskalleos:


stejskalleos:
@nadjaheitmann Latest Foreman is supporting EL9 only, so we don't need it anymore.

stejskalleos:
Failing tests are fixed in https://github.com/theforeman/foreman_puppet/pull/414

nadjaheitmann (commented):
I think you can also remove `module_hotfixes: true` . It is not needed for EL9.

PR #419, opened 2025-03-05
Use trusted publishers to release the gem

ekohl:


nadjaheitmann:
Looks like CI has been broken for a while. Guess we should fix that.

ekohl:
* Created Rubygems trusted publisher
* Removed old API key on both GH and Rubygems

PR #423, opened 2025-05-14
Update integration tests for jquery3

nadjaheitmann:
Needed because of https://github.com/theforeman/foreman/commit/871a647ddd17b163f9f8f1c18b2bcd4f492884f7

PR #424, opened 2025-05-14
Update integration tests for jquery3 part 2

nadjaheitmann:
For some reason, the integration tests were not run on the second push in https://github.com/theforeman/foreman_puppet/pull/423, which we realized only after merging. So here is a fixup PR.

### DIFF
--- a/test/integration/foreman_puppet/host_js_test.rb
+++ b/test/integration/foreman_puppet/host_js_test.rb
@@ -40,14 +40,11 @@ module ForemanPuppet
         select2 'Location 1', from: 'host_location_id'
         wait_for_ajax
         select2 env1.name, from: 'host_puppet_attributes_environment_id'
-        wait_for_ajax
         select2 hg.name, from: 'host_hostgroup_id'
         wait_for_ajax
 
         click_link 'Operating System'
-        wait_for_ajax
         select2 os.architectures.first.name, from: 'host_architecture_id'
-        wait_for_ajax
         select2 os.title, from: 'host_operatingsystem_id'
         uncheck('host_build')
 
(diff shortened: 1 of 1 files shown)


nadjaheitmann:
@MariaAga Do you have any idea why the integration tests are failing and how to fix it? https://github.com/theforeman/foreman_puppet/actions/runs/15041967945/job/42275651182?pr=424#step:21:30 

We tried to debug it but could not reproduce the error locally.

MariaAga:
This took awhile, so recording the debugging:
in foreman core, application.js, to find which tooltip is even hiding the select:
``` js
function onContentLoad() {
const targetNode = document.querySelector('body');
const config = { attributes: true, childList: true, subtree: true };
const callback = (mutationList, observer) => {
  for (const mutation of mutationList) {
    if (mutation.type === "childList") {
      mutation.addedNodes.forEach( n =>
      {
        if (n?.classList?.length && [...n.classList].includes('tooltip')){
          console.log('TOOLTIPHERE')
           console.log(n.innerHTML)
        }
      }
      )
  }
};

const observer = new MutationObserver(callback);
observer.observe(targetNode, config);
```

Then I ran `bundle exec rake test:foreman_puppet TEST=../foreman_puppet/test/integration/foreman_puppet/host_js_test.rb TESTOPTS="--name=/inherit/" SHOW_JS_LOG=1`
and it logged the env select choices

I tried clicking in random places before and after the OS tab change and nothing helped.
I decided to remove the tooltips here: https://github.com/theforeman/foreman/pull/10546

BTW:
I tried to reproduce it locally and never saw the tooltip not being hidden, but I did see an error when selecting the puppet env in the host form in this function: updatePuppetclasses
so looks like we need to change the line in this function to have a condition: ` if (url?.match('hostgroups')) {`

nadjaheitmann:
Thanks  @MariaAga , I had the PR running against your PR and it runs fine now.

adamruzicka:
@nadjaheitmann The PR in foreman was merged, could you please drop 5364155 ?

adamruzicka:
The test failures were not introduced here, opened #425 to address that

adamruzicka:
Thank you @nadjaheitmann !

PR #425, opened 2025-07-08
Fix js lints

adamruzicka:


adamruzicka:
And we're green

adamruzicka:
Thank you @ofedoren for keeping me on the right track

PR #429, opened 2025-09-24
Add Puppet proxy (CA) bulk actions

nadjaheitmann:
@adamruzicka @jeremylenz I have created a draft for the Puppet Proxy update bulk action. It is not more than a draft and I know there is a bunch of things missing (e.g. clearing proxy, CA proxy, tests, ...). I basically took code from Foreman and Katello and mixed it into this. Do you mind checking whether the general approach is the direction we want to go. UI wise, I think this is straight forward, but I was not sure whether 
- extending the BulkHostManager
-  adding a new API endpoint to the HostsBulkActionsController
- adding the API endpoint itself to the Foreman rather than the foreman_puppet namespace
- 
is the way to go. 

I appreciate your opinions.

### DIFF
--- a/app/controllers/foreman_puppet/api/v2/hosts_bulk_actions_controller.rb
+++ b/app/controllers/foreman_puppet/api/v2/hosts_bulk_actions_controller.rb
@@ -1,0 +1,102 @@
+module ForemanPuppet
+  module Api
+    module V2
+      class HostsBulkActionsController < ::ForemanPuppet::Api::V2::PuppetBaseController
+        include ::Api::V2::BulkHostsExtension
+        before_action :find_editable_hosts, only: %i[change_puppet_proxy remove_puppet_proxy]
+        before_action :find_smart_proxy, only: %i[change_puppet_proxy]
+
+        def_param_group :bulk_params do
+          param :organization_id, :number, required: true, desc: N_('ID of the organization')
+          param :included, Hash, required: true, action_aware: true do
+            param :search, String, required: false, desc: N_('Search string for hosts to perform an action on')
+            param :ids, Array, required: false, desc: N_('List of host ids to perform an action on')
+          end
+          param :excluded, Hash, required: true, action_aware: true do
+            param :ids, Array, required: false, desc: N_('List of host ids to exclude and not run an action on')
+          end
+        end
+
+        api :PUT, '/hosts/bulk/change_puppet_proxy', N_('Change Puppet (CA) Proxy')
+        param_group :bulk_params
+        param :proxy_id, :number, required: true, desc: N_('ID of the Puppet proxy to reassign the hosts to')
+        param :ca_proxy, :bool, required: true, desc: N_('True, if Puppet CA proxy should be changed instead of the Puppet proxy')
+        def change_puppet_proxy
+          error_hosts = ::BulkHostsManager.new(hosts: @hosts).change_puppet_proxy(@proxy, ca_proxy?)
+          process_bulk_puppet_proxy_response(
+            error_hosts,
+            success_message: format(n_(
+              'Updated host: changed %{proxy_type}',
+              'Updated hosts: changed %{proxy_type}',
+              @hosts.count
+            ), proxy_type: proxy_type),
+            error_message: format(n_(
+              'Failed to change %{proxy_type} for %{count} host',
+              'Failed to change %{proxy_type} for %{count} hosts',
+              error_hosts.count
+            ), proxy_type: proxy_type, count: error_hosts.count)
+          )
+        end
+
+        api :PUT, '/hosts/bulk/remove_puppet_proxy', N_('Remove Puppet (CA) Proxy')
+        param_group :bulk_params
+        param :ca_proxy, :bool, required: true, desc: N_('True, if Puppet CA proxy should be removed instead of the Puppet proxy')
+        def remove_puppet_proxy
+          error_hosts = ::BulkHostsManager.new(hosts: @hosts).change_puppet_proxy(nil, ca_proxy?)
+          process_bulk_puppet_proxy_response(
+            error_hosts,
+            success_message: format(n_(
+              'Updated host: removed %{proxy_type}',
+              'Updated hosts: removed %{proxy_type}',
+              @hosts.count
+            ), proxy_type: proxy_type),
+            error_message: format(n_(
+              'Failed to remove %{proxy_type} for %{count} host',
+              'Failed to remove %{proxy_type} for %{count} hosts',
+              error_hosts.count
+            ), proxy_type: proxy_type, count: error_hosts.count)
+          )
+        end
+
+        private
+
+        def find_editable_hosts
+          find_bulk_hosts(:edit_hosts, params)
+        end
+
+        def process_bulk_puppet_proxy_response(error_hosts, success_message:, error_message:)
+          if error_hosts.empty?
+            process_response(true, { message: success_message })
+          else
+            render_error(:bulk_hosts_error, status: :unprocessable_entity,
+              locals: { message: error_message, failed_host_ids: error_hosts })
+          end
+        end
+
+        def find_smart_proxy
+          feature = ca_proxy? ? 'Puppet CA' : 'Puppet'
+          @proxy = SmartProxy.with_features(feature).find_by(id: params[:proxy_id])
+
+          if @proxy.nil?
+            render json: {
+              error: {
+                message: format(_('A Smart Proxy with id %{id} and the %{proxy_type} feature could not be found.'), id: params[:proxy_id], proxy_type: proxy_type),
+              },
+            }, status: :unprocessable_entity
+            false
+          else
+            true
+          end
+        end
+
+        def ca_proxy?
+          Foreman::Cast.to_bool(params[:ca_proxy])
+        end
+
+        def proxy_type
+          ca_proxy? ? _('Puppet CA proxy') : _('Puppet proxy')
+        end
+      end
+    end
+  end
+end
--- a/app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
+++ b/app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
@@ -1,0 +1,24 @@
+module ForemanPuppet
+  module Extensions
+    module BulkHostsManager
+      extend ActiveSupport::Concern
+
+      def change_puppet_proxy(proxy, is_ca_proxy)
+        error_hosts = []
+        @hosts.each do |host|
+          if is_ca_proxy
+            host.puppet_ca_proxy = proxy
+          else
+            host.puppet_proxy = proxy
+          end
+          host.save(validate: false)
+        rescue StandardError => e
+          message = format(_('Failed to set proxy for %{host}.'), host: host)
+          Foreman::Logging.exception(message, e)
+          error_hosts << host.id
+        end
+        error_hosts
+      end
+    end
+  end
+end
--- a/config/api_routes.rb
+++ b/config/api_routes.rb
@@ -1,3 +1,14 @@
+Foreman::Application.routes.draw do
+  scope module: 'foreman_puppet' do
+    namespace :api, defaults: { format: 'json' } do
+      scope '(:apiv)', module: :v2, defaults: { apiv: 'v2' }, apiv: /v1|v2/, constraints: ApiConstraints.new(version: 2, default: true) do
+        match 'hosts/bulk/change_puppet_proxy', to: 'hosts_bulk_actions#change_puppet_proxy', via: [:put]
+        match 'hosts/bulk/remove_puppet_proxy', to: 'hosts_bulk_actions#remove_puppet_proxy', via: [:put]
+      end
+    end
+  end
+end
+
 ForemanPuppet::Engine.routes.draw do
   namespace :api, defaults: { format: 'json' } do
     scope '(:apiv)', module: :v2, defaults: { apiv: 'v2' }, apiv: /v1|v2/, constraints: ApiConstraints.new(version: 2, default: true) do
--- a/webpack/src/Extends/Hosts/ActionsBar/index.js
+++ b/webpack/src/Extends/Hosts/ActionsBar/index.js
@@ -1,0 +1,73 @@
+import React, { useContext } from 'react';
+import { Menu, MenuItem, MenuContent, MenuList } from '@patternfly/react-core';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { ForemanHostsIndexActionsBarContext } from 'foremanReact/components/HostsIndex';
+import { openBulkModal } from 'foremanReact/common/BulkModalStateHelper';
+import './ActionsBar.scss';
+
+const HostActionsBar = () => {
+  const { selectedCount, setMenuOpen } = useContext(
+    ForemanHostsIndexActionsBarContext
+  );
+
+  const handleOpenBulkModal = modalId => {
+    setMenuOpen(false);
+    setTimeout(() => openBulkModal(modalId, true), 0);
+  };
+
+  return (
+    <MenuItem
+      itemId="content-flyout-item"
+      key="content-flyout"
+      isDisabled={selectedCount === 0}
+      flyoutMenu={
+        <Menu ouiaId="content-flyout-menu" onSelect={() => setMenuOpen(false)}>
+          <MenuContent>
+            <MenuList>
+              <MenuItem
+                itemId="bulk-change-puppet-proxy-menu-item"
+                key="bulk-change-puppet-proxy-menu-item"
+                onClick={() => handleOpenBulkModal('bulk-change-puppet-proxy')}
+                isDisabled={selectedCount === 0}
+              >
+                {__('Change Puppet proxy')}
+              </MenuItem>
+              <MenuItem
+                itemId="bulk-remove-puppet-proxy-menu-item"
+                key="bulk-remove-puppet-proxy-menu-item"
+                onClick={() => handleOpenBulkModal('bulk-remove-puppet-proxy')}
+                isDisabled={selectedCount === 0}
+              >
+                {__('Remove Puppet proxy')}
+              </MenuItem>
+              <MenuItem
+                itemId="bulk-change-puppet-ca-proxy-menu-item"
+                key="bulk-change-puppet-ca-proxy-menu-item"
+                onClick={() =>
+                  handleOpenBulkModal('bulk-change-puppet-ca-proxy')
+                }
+                isDisabled={selectedCount === 0}
+              >
+                {__('Change Puppet CA proxy')}
+              </MenuItem>
+              <MenuItem
+                itemId="bulk-remove-puppet-ca-proxy-menu-item"
+                key="bulk-remove-puppet-ca-proxy-menu-item"
+                onClick={() =>
+                  handleOpenBulkModal('bulk-remove-puppet-ca-proxy')
+                }
+                isDisabled={selectedCount === 0}
+              >
+                {__('Remove Puppet CA proxy')}
+              </MenuItem>
+            </MenuList>
+          </MenuContent>
+        </Menu>
+      }
+    >
+      {__('Puppet')}
+    </MenuItem>
+  );
+};
+
+export default HostActionsBar;
--- a/webpack/src/Extends/Hosts/BulkActions/BulkChangePuppetCAProxy/index.js
+++ b/webpack/src/Extends/Hosts/BulkActions/BulkChangePuppetCAProxy/index.js
@@ -1,0 +1,40 @@
+import React, { useContext } from 'react';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { ForemanActionsBarContext } from 'foremanReact/components/HostDetails/ActionsBar';
+import { useBulkModalOpen } from 'foremanReact/common/BulkModalStateHelper';
+
+import BulkChangeProxyCommon from '../BulkChangeProxyCommon';
+
+const BulkChangePuppetCAProxyScene = () => {
+  const {
+    selectAllHostsMode,
+    selectedCount,
+    selectedResults,
+    fetchBulkParams,
+  } = useContext(ForemanActionsBarContext);
+  const { isOpen, close: closeModal } = useBulkModalOpen(
+    'bulk-change-puppet-ca-proxy'
+  );
+  return (
+    <BulkChangeProxyCommon
+      isCAProxy
+      fetchBulkParams={fetchBulkParams}
+      selectedCount={selectedCount}
+      selectedResults={selectedResults}
+      selectAllHostsMode={selectAllHostsMode}
+      isOpen={isOpen}
+      closeModal={closeModal}
+      selectMessage={__('Select a Puppet CA Proxy')}
+      handleErrorMessage={__('Failed to change Puppet CA Proxy')}
+      changeMessage={__('Change Puppet CA Proxy')}
+      allHostsMessage={__(
+        'Changing the Puppet CA proxy will affect {boldCount} selected hosts. Some hosts may already have been associated with the selected Puppet CA proxy.'
+      )}
+      someHostsMessage={__(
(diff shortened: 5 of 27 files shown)


app/controllers/concerns/foreman_puppet/extensions/api_v2_hosts_bulk_actions_controller.rb
```diff
@@ -0,0 +1,20 @@
+module ForemanPuppet
+  module Extensions
+    module ApiV2HostsBulkActionsController
+      extend ActiveSupport::Concern
+
+      included do
+        before_action :find_editable_hosts, :only => [:change_puppet_proxy]
```
adamruzicka: Won't this override the before action from the base class in https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/hosts_bulk_actions_controller.rb#L8 ?

In this case I'd prefer katello's approach - have a [brand new controller](https://github.com/Katello/katello/blob/8f4da2816a1a44f99dfda030056fe28b53978ddd/app/controllers/katello/api/v2/hosts_bulk_actions_controller.rb) rather than trying to extend the existing one.

app/controllers/concerns/foreman_puppet/extensions/api_v2_hosts_bulk_actions_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+
+        api :PUT, "/hosts/bulk/change_puppet_proxy", N_("Change Puppet Proxy")
+        param_group :bulk_host_ids
+        param :proxy_id, :number, :required => true, :desc => N_("ID of the Puppet proxy to reassign the hosts to")
```
adamruzicka: This being required means it can be used to reassign the host to a different proxy, but this cannot be used to unassign it, correct?

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
```diff
@@ -0,0 +1,26 @@
...
+      def change_puppet_proxy(proxy_id, ca_proxy)
+        proxy = SmartProxy.find_by(id: proxy_id)
+        begin
+          @hosts.each do |host|
+            if ca_proxy
+              host.puppet_ca_proxy = proxy
+            else
+              host.puppet_proxy = proxy
+            end
+            host.save(:validate => false)
```
adamruzicka: Do we need callbacks to be run or could we take a shortcut and assign the proxy with a single update query?

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
```diff
@@ -0,0 +1,26 @@
...
+        rescue StandardError => e
+            message = format(_('Failed to set %{proxy_type} proxy for %{host}.'), host: host, proxy_type: proxy_type)
+            Foreman::Logging.exception(message, e)
+        end
+        rescue RecordNotFoundError => e
+          Foreman::Logging.exception(_('Cound not find Smart Proxy'))
+        end
```
adamruzicka, line 19:
```suggestion
        rescue StandardError => e
            message = format(_('Failed to set %{proxy_type} proxy for %{host}.'), host: host, proxy_type: proxy_type)
            Foreman::Logging.exception(message, e)
        end
        rescue ActiveRecord::RecordNotFound => e
          Foreman::Logging.exception(_('Cound not find Smart Proxy'))
        end
```

And even then, ActiveRecord::RecordNotFound is a subclass of StandardError so the record not found branch would never match

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
```diff
@@ -0,0 +1,26 @@
+module ForemanPuppet
+  module Extensions
+    module BulkHostsManager
+      extend ActiveSupport::Concern
+
+      def change_puppet_proxy(proxy_id, ca_proxy)
+        proxy = SmartProxy.find_by(id: proxy_id)
```
adamruzicka: This just returns nil if no proxy was found. Is that expected?

adamruzicka (commented):
Left some notes inline, can't speak much to the frontend

app/controllers/concerns/foreman_puppet/extensions/api_v2_hosts_bulk_actions_controller.rb
nadjaheitmann: Yeah, it does. Is there a way not to overwrite it? If not, it has to be a new controller, indeed.

app/controllers/concerns/foreman_puppet/extensions/api_v2_hosts_bulk_actions_controller.rb
nadjaheitmann: Correct. I have not yet implemented the unassign method, so the implementation details about that one are still open.

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
nadjaheitmann: Yes.  I would solve that depending on how I implement the unassign method. The PR is still very drafty.

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
nadjaheitmann: Not sure, if we need callbacks to be honest. The original action does run some validations about validating the proxy I think: https://github.com/theforeman/foreman_puppet/blob/master/app/controllers/concerns/foreman_puppet/extensions/hosts_controller_extensions.rb#L143

Can the update command also take the `validate => false` parameter?

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
nadjaheitmann: Yeah, that one is messy - this also somehow depends if we want a separate method for unassign.

nadjaheitmann (commented):
Thanks @adamruzicka !

app/controllers/concerns/foreman_puppet/extensions/api_v2_hosts_bulk_actions_controller.rb
adamruzicka: From the top of my head I can't suggest any clean way that would make it behave the way we need. Hacks could be done, sure, with the cleanest one probably being just calling it directly in `#change_puppet_proxy` without relying on controller callbacks.

webpack/src/Extends/Hosts/ActionsBar/index.js
```diff
@@ -0,0 +1,105 @@
+import React, { useContext, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { useDispatch } from 'react-redux';
+import { Menu, MenuItem, MenuContent, MenuList } from '@patternfly/react-core';
+import { BanIcon } from '@patternfly/react-icons';
+import { translate as __ } from 'foremanReact/common/I18n';
+import { ForemanHostsIndexActionsBarContext } from 'foremanReact/components/HostsIndex';
+import { useForemanModal } from 'foremanReact/components/ForemanModal/ForemanModalHooks';
```
jeremylenz: `useForemanModal` is being removed, so using it here is probably not a good idea. In Katello we had to do some [kind of ugly hacks](https://github.com/Katello/katello/pull/11626), too, so maybe worth consulting with @Lukshio about the best way to go about it.

webpack/src/Extends/Hosts/ActionsBar/index.js
```diff
@@ -0,0 +1,105 @@
...
+        <Menu ouiaId="content-flyout-menu" onSelect={() => setMenuOpen(false)}>
+          <MenuContent>
+            <MenuList>
+              <MenuItem
+                itemId="bulk-change-puppet-proxy-menu-item"
+                key="bulk-change-puppet-proxy-menu-item"
+                onClick={openBulkChangePuppetProxy}
+                isDisabled={selectedCount === 0}
+              >
+                {__('Change Puppet Proxy')}
```
jeremylenz:
```suggestion
                {__('Change Puppet proxy')}
```

As a rule we use sentence case in everything except page titles.

webpack/src/Extends/Hosts/ActionsBar/index.js
```diff
@@ -0,0 +1,105 @@
...
+              </MenuItem>
+              <MenuItem
+                itemId="bulk-remove-puppet-proxy-menu-item"
+                key="bulk-remove-puppet-proxy-menu-item"
+                onClick={openBulkRemovePuppetProxy}
+                isDisabled={selectedCount === 0}
+              >
+                {__('Remove Puppet Proxy')}
```
jeremylenz:
```suggestion
                {__('Remove Puppet proxy')}
```

webpack/src/Extends/Hosts/ActionsBar/index.js
```diff
@@ -0,0 +1,105 @@
...
+              >
+                {__('Change Puppet CA Proxy')}
+              </MenuItem>
+              <MenuItem
+                itemId="bulk-remove-puppet-ca-proxy-menu-item"
+                key="bulk-remove-puppet-ca-proxy-menu-item"
+                onClick={openBulkRemovePuppetCAProxy}
+                isDisabled={selectedCount === 0}
+              >
+                {__('Remove Puppet CA Proxy')}
```
jeremylenz:
```suggestion
                {__('Remove Puppet CA proxy')}
```

jeremylenz (commented):
Did a quick read of the frontend code, a couple comments below

nadjaheitmann:
This PR uses https://github.com/theforeman/foreman/commit/40938e445976db7987b965352c2025d2401bc479 , so it will only work with Foreman 3.19+

app/controllers/concerns/foreman_puppet/extensions/api_v2_hosts_bulk_actions_controller.rb
nadjaheitmann: I changed that one and implemented a new controller.

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
nadjaheitmann: Reworked that part.

nadjaheitmann:
@jeremylenz @adamruzicka Can you please re-review the PR? It is not completely in line with what the Bulk actions in Foreman core do/return but it could be implemented in a follow-up. (it does not return a nice link to the failed hosts)

image

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
nadjaheitmann: re-worked

jeremylenz (approved):
JS changes LGTM, did not test. 👍

app/controllers/foreman_puppet/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+          if error_hosts.empty?
+            process_response(true, { message: success_message })
+          else
+            render_error(:bulk_hosts_error, status: :unprocessable_entity,
+              locals: { message: error_message, failed_host_ids: error_hosts })
+          end
+        end
+
+        def find_smart_proxy
+          @proxy = SmartProxy.find_by(id: params[:proxy_id])
```
adamruzicka: This should probably also check that the proxy has the right features

webpack/src/Extends/Hosts/BulkActions/BulkChangePuppetCAProxy/index.js
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <BulkChangeProxyCommom
+      isCAProxy
+      fetchBulkParams={fetchBulkParams}
+      selectedCount={selectedCount}
+      selectedResults={selectedResults}
+      selectAllHostsMode={selectAllHostsMode}
+      isOpen={isOpen}
+      closeModal={closeModal}
+      title={__('Change Puppet CA Proxy')}
```
adamruzicka: I don't see the BulkChangeProxyCommon accepting a title prop?

webpack/src/Extends/Hosts/BulkActions/BulkChangePuppetCAProxy/index.js
```diff
@@ -0,0 +1,41 @@
...
+    selectAllHostsMode,
+    selectedCount,
+    selectedResults,
+    fetchBulkParams,
+  } = useContext(ForemanActionsBarContext);
+  const { isOpen, close: closeModal } = useBulkModalOpen(
+    'bulk-change-puppet-ca-proxy'
+  );
+  return (
+    <BulkChangeProxyCommom
```
adamruzicka:
```suggestion
    <BulkChangeProxyCommon
```

webpack/src/Extends/Hosts/BulkActions/BulkChangePuppetProxy/index.js
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <BulkChangeProxyCommon
+      isCAProxy={false}
+      fetchBulkParams={fetchBulkParams}
+      selectedCount={selectedCount}
+      selectedResults={selectedResults}
+      selectAllHostsMode={selectAllHostsMode}
+      isOpen={isOpen}
+      closeModal={closeModal}
+      title={__('Change Puppet Proxy')}
```
adamruzicka: See above

adamruzicka (changes requested):

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
```diff
@@ -0,0 +1,24 @@
...
+          if ca_proxy
+            host.puppet_ca_proxy = proxy
+          else
+            host.puppet_proxy = proxy
+          end
+          host.save(validate: false)
+        rescue StandardError => e
+          message = format(_('Failed to set proxy for %{host}.'), host: host)
+          Foreman::Logging.exception(message, e)
+          error_hosts << host.id
```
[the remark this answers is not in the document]
adamruzicka: One could argue that with `validate: false` this shouldn't really happen?

app/controllers/foreman_puppet/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+          end
+        end
+
+        api :PUT, '/hosts/bulk/change_puppet_proxy', N_('Change Puppet (CA) Proxy')
+        param_group :bulk_params
+        param :proxy_id, :number, required: true, desc: N_('ID of the Puppet proxy to reassign the hosts to')
+        param :ca_proxy, :bool, required: true, desc: N_('True, if Puppet CA proxy should be changed instead of the Puppet proxy')
+        def change_puppet_proxy
+          error_hosts = ::BulkHostsManager.new(hosts: @hosts).change_puppet_proxy(@proxy, params[:ca_proxy])
+          process_bulk_puppet_proxy_response(
```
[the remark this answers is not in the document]
adamruzicka: Foreman has a helper for string->bool casting

webpack/src/Extends/Hosts/BulkActions/BulkChangePuppetCAProxy/index.js
nadjaheitmann: Nice catch, I wonder how this did not show up in my tests.

config/api_routes.rb
```diff
@@ -1,3 +1,14 @@
...
+        match 'hosts/bulk/change_puppet_proxy', to: 'hosts_bulk_actions#change_puppet_proxy', via: [:put]
+        match 'hosts/bulk/remove_puppet_proxy', to: 'hosts_bulk_actions#remove_puppet_proxy', via: [:put]
+      end
+    end
+  end
+end
+
 ForemanPuppet::Engine.routes.draw do
   namespace :api, defaults: { format: 'json' } do
     scope '(:apiv)', module: :v2, defaults: { apiv: 'v2' }, apiv: /v1|v2/, constraints: ApiConstraints.new(version: 2, default: true) do
```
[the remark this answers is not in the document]
nadjaheitmann: The routes are on purpose still in the Foreman API. This is on purpose as the whole Puppet Proxy thing is still part of Foreman core actually. Any opinion on this @adamruzicka ?

PS: I fixed the UI issue that copilot mentions here.

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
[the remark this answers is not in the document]
nadjaheitmann: I would assume it just works. Not sure what would happen if you e.g. assign a proxy and it is deleted between the check in the API and call the to `host.save`.

webpack/src/Extends/Hosts/BulkActions/BulkChangePuppetCAProxy/index.js
nadjaheitmann: It does not need one because title is the same as the `changeMessage`. I remove it for now.

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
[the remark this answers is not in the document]
nadjaheitmann: @adamruzicka Any preference what to do with this code? The validation is in line with the original implementation: https://github.com/theforeman/foreman_puppet/blob/master/app/controllers/concerns/foreman_puppet/extensions/hosts_controller_extensions.rb#L73 

Not sure what would happen if we change the behavior now. I agree that having the error catching does not make much sense from this perspective. However, removing it also feels wrong somehow. 

Can you please take another look? I would like to finish this up and it just keeps growing and growing :)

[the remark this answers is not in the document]
adamruzicka: I'd be fine with keeping it as it was. Sometimes copilot gets overly zealous

app/controllers/foreman_puppet/api/v2/hosts_bulk_actions_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+                (!ca_proxy? && !@proxy.has_feature?('Puppet'))
+            render json: {
+              error: {
+                message: format(_('The Smart Proxy you have provided does not have the feature %{proxy_type}.'), proxy_type: proxy_type),
+              },
+            }, status: :unprocessable_entity
+            false
+          else
+            true
+          end
```
adamruzicka: _personal preference, feel free to ignore:_
```suggestion
          feature = ca_proxy? ? 'Puppet CA' : 'Puppet'
          @proxy = SmartProxy.with_features(feature).find_by(id: params[:proxy_id])

          if @proxy.nil?
            render json: {
              error: {
                message: format(_('The %{proxy_type} with id %{id} could not be found.'), proxy_type: proxy_type, id: params[:proxy_id]),
              },
            }, status: :unprocessable_entity
            false
          else
            true
          end
```

config/api_routes.rb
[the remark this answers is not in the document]
adamruzicka: > This is on purpose as the whole Puppet Proxy thing is still part of Foreman core actually.

This is a bit unfortunate, but that's where we are.

adamruzicka (commented):
I currently don't have an environment to test this, but code wise it looks good.

app/controllers/foreman_puppet/api/v2/hosts_bulk_actions_controller.rb
nadjaheitmann: I admit that the code could be a bit more elegant but I prefer to tell the user that the proxy is actually missing the feature.

config/api_routes.rb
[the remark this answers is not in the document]
nadjaheitmann: It is and it will be a whole load of work to change that I guess.

nadjaheitmann:
> environ

Thanks @adamruzicka . I accidentally requested another review from you, feel free to ignore that.

nadjaheitmann:
> I currently don't have an environment to test this, but code wise it looks good.

I had a test session with @sbernhard and @Manisha15 . We tested the PR, verified that the permissions work and added a UI warning in case there is no smart proxy with Puppet(CA) feature available:

image
image

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
```diff
@@ -0,0 +1,24 @@
+module ForemanPuppet
+  module Extensions
+    module BulkHostsManager
+      extend ActiveSupport::Concern
+
+      def change_puppet_proxy(proxy, ca_proxy)
```
sbernhard:
```suggestion
      def change_puppet_proxy(proxy, is_ca_proxy?)
```

webpack/src/Extends/Hosts/ActionsBar/index.js
```diff
@@ -0,0 +1,73 @@
...
+              </MenuItem>
+              <MenuItem
+                itemId="bulk-change-puppet-ca-proxy-menu-item"
+                key="bulk-change-puppet-ca-proxy-menu-item"
+                onClick={() =>
+                  handleOpenBulkModal('bulk-change-puppet-ca-proxy')
+                }
+                isDisabled={selectedCount === 0}
+              >
+                {__('Change Puppet CA proxy')}
```
sbernhard, line 51: In old UI, it was just "Change Puppet CA" - in the documentatin its mentioning "Puppet CA Smart Proxy"
a) Change Puppet CA
b) Change Puppet CA Smart Proxy

webpack/src/Extends/Hosts/ActionsBar/index.js
```diff
@@ -0,0 +1,73 @@
...
+        <Menu ouiaId="content-flyout-menu" onSelect={() => setMenuOpen(false)}>
+          <MenuContent>
+            <MenuList>
+              <MenuItem
+                itemId="bulk-change-puppet-proxy-menu-item"
+                key="bulk-change-puppet-proxy-menu-item"
+                onClick={() => handleOpenBulkModal('bulk-change-puppet-proxy')}
+                isDisabled={selectedCount === 0}
+              >
+                {__('Change Puppet proxy')}
```
sbernhard, line 33: In old UI it was "Change Puppet Master". In documentation its "Puppet Smart Proxy"
a) Change Puppet Master
b) Change Puppet Smart Proxy

webpack/src/Extends/Hosts/ActionsBar/index.js
nadjaheitmann: On the 'Edit' page, it is like this:
image

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
nadjaheitmann: Changed. At least without the '?'

webpack/src/Extends/Hosts/ActionsBar/index.js
nadjaheitmann: We decided to keep it as is.

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
```diff
@@ -0,0 +1,24 @@
+module ForemanPuppet
+  module Extensions
+    module BulkHostsManager
+      extend ActiveSupport::Concern
+
+      def change_puppet_proxy(proxy, is_ca_proxy)
```
sbernhard, line 6:
```suggestion
      def change_puppet_proxy(proxy, is_ca_proxy?)
```

sbernhard (approved):

app/services/concerns/foreman_puppet/extensions/bulk_hosts_manager.rb
nadjaheitmann: Does not work, this is a parameter and they cannot have '?'

Manisha15 (approved):
LGTM

PR #430, opened 2025-09-25
Add breaks-robottelo workflow as per theforeman/foreman PR 10568

lhellebr:
Before merging, `breaks-robottelo` label needs to be added to the repo manually by a maintainer

### DIFF
--- a/.github/workflows/test-breaking.yaml
+++ b/.github/workflows/test-breaking.yaml
@@ -1,0 +1,15 @@
+name: Label a PR `breaks-robottelo` on appropriate comment
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  breaks-robottelo:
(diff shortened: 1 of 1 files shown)


nadjaheitmann:
@lhellebr Created label: https://github.com/theforeman/foreman_puppet/issues?q=state%3Aopen%20label%3Abreaks-robottelo 

This is a cool idea, I hope it will help you! Greatest hurdle is getting this in the heads of developers I suppose.

Some years ago, we tried to establish the `breaking change` label in Foreman that indicates which changes would cause a plugin to break such that plugin maintainers could easily adjust. But that did not survive, unfortunately.

ofedoren (approved):
Thanks, @lhellebr and @nadjaheitmann !

PR #434, opened 2025-10-02
Use optional chaining for URL match check

stejskalleos:


MariaAga:
I'm ok with this as a quick fix, but want to point to a bigger issue I described on Matrix :
I checked some things, and generally it seems like foreman_puppet has issues on the hosts page. anytime hostgroup or env is changed, it calls the updatePuppetclasses function, but the html it searches for only exists in these conditions:
host_onlyif = ->(host, context) { context.send(:accessible_resource, host, :smart_proxy, :name, association: :puppet_proxy).present? }
I'm not sure if it will be just enough to add a condition to check in updatePuppetclasses if puppet tab is even loaded

stejskalleos:
> I'm not sure if it will be just enough to add a condition to check in updatePuppetclasses if puppet tab is even loaded

Well, that I would leave up to the foreman_puppet maintainers :)

MariaAga (approved):
Since its not breaking anything after this workaround I'll leave it be.
Thanks!

PR #439, opened 2025-12-21
Show the view state of puppet modules/classes using chevrons on puppetclass selection page

me-minus:
This commit changes the add puppet class view to be a bit more like a tree by changing the big + to be a right chevron or down chevron if the module has been opened.

When I have shown people how to create machines with the help of foreman, many have wondered what the difference between the '+' symbols when adding puppet classes are. 

Original (from foreman 3.19.0 + foreman_puppet 10.1.1):
Screenshot From 2026-06-20 15-35-25

Changed version:
Screenshot From 2026-06-20 21-57-30

### DIFF
--- a/app/views/foreman_puppet/puppetclasses/_classes.html.erb
+++ b/app/views/foreman_puppet/puppetclasses/_classes.html.erb
@@ -2,21 +2,36 @@
   <div class="col-md-6 classes">
     <% group.each do |list| %>
       <% next if list.nil? %>
-      <ul class="puppetclass_group">
-        <li><%= puppetclass_group_with_icon(list, selected_puppet_classes) %>
-          <ul id="pc_<%= list.first %>" style="display: none;">
-            <% for klass in list.last.sort %>
-              <% unless authorized_for(controller: :host_editing, action: :edit_classes) %>
-                <li data-original-title="<%= _('Not authorized to edit classes') %>" rel="twipsy"><%= h klass.name %></li>
-              <% else %>
-                <li id="puppetclass_<%= klass.id %>" class="puppetclass<%= ' hide' if selected_puppet_classes.include?(klass) %>">
-                  <%= link_to_add_puppetclass(klass, resource_type) %>
-                </li>
-              <% end %>
-            <% end %>
-          </ul>
-        </li>
-      </ul>
+
+      <div class="panel-group paneless puppetclass_group" id="<%= list.first %>-avail-classes" style="margin-bottom: 1px;">
+	<div class="panel-heading" style="padding-top: 2px; padding-bottom: 2px;">
+	  <p class="panel-title">
+	    <a data-toggle="collapse" data-parent="#<%= list.first %>-avail-classes" href="#<%= list.first %>-avail-classes-list" class="collapsed">
+	      <%= h list.first %>
+	    </a>
+	  </p>
+	</div>
+	<div id="<%= list.first %>-avail-classes-list" class="panel-collapse collapse">
+	  <div class="panel-body">
+	    <ul style="margin-bottom: 1px;">
+	      <% for klass in list.last.sort %>
+	        <% unless authorized_for(:controller => :host_editing, :action => :edit_classes) %>
+	          <li data-original-title="<%= _('Not authorized to edit classes') %>" rel="twipsy">
+	            <%= h klass.name %>
+		  </li>
+		<% else %>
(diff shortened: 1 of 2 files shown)


nadjaheitmann:
Hi @me-minus ! Thanks for the contribution! Do you mind providing a screenshot that show cases your change?

me-minus:
foreman_with_chevrons

nadjaheitmann:
Looks good to me and works fine. But I don't want to merge this without the consent of @adamruzicka .

app/views/foreman_puppet/puppetclasses/_classes.html.erb
```diff
@@ -2,21 +2,36 @@
...
-                  <%= link_to_add_puppetclass(klass, resource_type) %>
-                </li>
-              <% end %>
-            <% end %>
-          </ul>
-        </li>
-      </ul>
+
+      <div class="panel-group paneless puppetclass_group" id="<%= list.first %>-avail-classes" style="margin-bottom: 1px;">
+	<div class="panel-heading" style="padding-top: 2px; padding-bottom: 2px;">
```
adamruzicka: I can't say I'm exactly happy about inline styling all over the place. Could that be extracted to classes?

app/views/foreman_puppet/puppetclasses/_classes.html.erb
```diff
@@ -2,21 +2,36 @@
...
+	        <% unless authorized_for(:controller => :host_editing, :action => :edit_classes) %>
+	          <li data-original-title="<%= _('Not authorized to edit classes') %>" rel="twipsy">
+	            <%= h klass.name %>
+		  </li>
+		<% else %>
+		  <% style = ["puppetclass",
+		              selected_puppet_classes.include?(klass) ? "hide" : "",
+		             ].join(' ') %>
+		  <%= content_tag(:li, klass, :id=>"puppetclass_#{klass.id}", :class=>style) do %>
+                    <%= link_to_add_puppetclass(klass,type) %>
```
adamruzicka:
```suggestion
		  <%= content_tag(:li, klass, :id => "puppetclass_#{klass.id}", :class => style) do %>
                    <%= link_to_add_puppetclass(klass, type) %>
```

webpack/src/foreman_class_edit.js
```diff
@@ -28,9 +28,18 @@ export function filterPuppetClasses(item) {
     classElems
       .hide()
       .has(`[data-class-name*="${term}"]`)
-      .show();
+      .show()
+      .filter('div')
+      .find('a.collapsed')
+      .attr('aria-expanded','true')
+      .click();
```
adamruzicka, line 35: Why the click?

webpack/src/foreman_class_edit.js
```diff
@@ -28,9 +28,18 @@ export function filterPuppetClasses(item) {
     classElems
       .hide()
       .has(`[data-class-name*="${term}"]`)
-      .show();
+      .show()
+      .filter('div')
+      .find('a.collapsed')
```
adamruzicka, line 33: Where is the `collapsed` class being set on a `<a>`?

adamruzicka (changes requested):
The erb is mixing tabs and spaces for indentation. Could you please only use spaces for that?

app/views/foreman_puppet/puppetclasses/_classes.html.erb
me-minus: fixed

webpack/src/foreman_class_edit.js
me-minus: To make the expansion/collapse work when changing the text in the searchbar

webpack/src/foreman_class_edit.js
me-minus: I don't really know. 
I used the developers pane to find out how the attributes changed with some hints from the _classes.html.erb file. 
The code was created by changing/playing with the left menu pane code in an older foreman version (probably before parternfly 5)

app/views/foreman_puppet/puppetclasses/_classes.html.erb
me-minus: I could not find a way to extract the styling, so I removed it but then the spacing gets wonky 
Screenshot From 2026-03-29 15-26-54
Do you have any hints on where I could look?

PR #440, opened 2026-03-05
Switch manual links to point to docs

adamruzicka:
Draft until I figure out a replacement for `4.2.5ParameterizedClasses` which doesn't seem to have a direct counterpart in the new docs.

### DIFF
--- a/app/views/foreman_puppet/config_groups/welcome.html.erb
+++ b/app/views/foreman_puppet/config_groups/welcome.html.erb
@@ -8,6 +8,6 @@
                                     :iconType => 'fa',
                                     :header => _('Config Groups'),
                                     :description => description,
-                                    :documentation => { :url => documentation_url("using-config-groups") },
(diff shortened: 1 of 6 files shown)


maximiliankolb:
Please note that as of now, docs.theforeman.org is only the official documentation for Foreman+Katello. As you can run foreman_puppet on Foreman only too, we'd need to deprecate Foreman Manual first, promote docs.theforeman.org to the official docs, and then merge this.

adamruzicka:
> As you can run foreman_puppet on Foreman only too, we'd need to deprecate Foreman Manual first, promote docs.theforeman.org to the official docs, and then merge this.

At the same time, the manual wasn't really updated to reflect the extraction of f-puppet away from foreman itself, right?

maximiliankolb:
> > As you can run foreman_puppet on Foreman only too, we'd need to deprecate Foreman Manual first, promote docs.theforeman.org to the official docs, and then merge this.
> 
> At the same time, the manual wasn't really updated to reflect the extraction of f-puppet away from foreman itself, right?

Yes, to my understanding, from a docs perspective, the Foreman Manual is less maintained and not used for any downstream products. I think that Foreman Manual is mostly updated if the installation of Foreman/Foreman+Katello changes.

PR #441, opened 2026-03-06
Fixes #39210 - Fix the tests

adamruzicka:
This most likely broke when https://github.com/theforeman/foreman/pull/10701 went in.

### DIFF
--- a/app/models/concerns/foreman_puppet/extensions/user.rb
+++ b/app/models/concerns/foreman_puppet/extensions/user.rb
@@ -9,19 +9,7 @@ module ForemanPuppet
 
       module PatchedMethods
         def visible_environments
-          authorized_scope = ForemanPuppet::Environment.unscoped.authorized(:view_environments)
-          authorized_scope = authorized_scope
-                             .joins(:taxable_taxonomies)
-                             .where('taxable_taxonomies.taxonomy_id' => taxonomy_ids[:organizations] + taxonomy_ids[:locations])
-          result = authorized_scope.distinct.pluck(:name)
-          if ::User.current.admin?
-            # Admin users can also see Environments that do not have any organization or location, even when
-            # organizations and locations are enabled.
-            untaxed_env_ids = TaxableTaxonomy.where(taxable_type: 'ForemanPuppet::Environment').distinct.select(:taxable_id)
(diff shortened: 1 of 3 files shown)


app/models/concerns/foreman_puppet/extensions/user.rb
```diff
@@ -9,19 +9,7 @@ module User
...
-          result = authorized_scope.distinct.pluck(:name)
-          if ::User.current.admin?
-            # Admin users can also see Environments that do not have any organization or location, even when
-            # organizations and locations are enabled.
-            untaxed_env_ids = TaxableTaxonomy.where(taxable_type: 'ForemanPuppet::Environment').distinct.select(:taxable_id)
-            untaxed_environments = ForemanPuppet::Environment.unscoped.where.not(id: untaxed_env_ids).pluck(:name)
-            result += untaxed_environments
-          end
-          result
+          ForemanPuppet::Environment.authorized(:view_environments).pluck(:name)
```
adamruzicka, line 12: This currently requires https://github.com/theforeman/foreman/pull/10898 . An alternative would be this suggestion

```suggestion
          ForemanPuppet::Environment.authorized(:view_environments).distinct.pluck(:name)
```

ofedoren (commented):
This should also bump required foreman version if we merge the change in core :/

ofedoren:
~I hope the tests will fail :)~ As expected, currently the tests fail.

And maybe it's worth to create a RM issue to link to the core's one? And if the tests fail, I guess it won't hurt to add a commit to run this CI with https://github.com/theforeman/foreman/pull/10898 to ensure the tests pass with that change.

Although this is pretty straightforward issue/change (thus not required), I'd like to make a pattern from testing with the dependency.

adamruzicka:
> And if the tests fail, I guess it won't hurt to add a commit to run this CI 

Yeah, I have it ready, I was just waiting for the tests to fail.

ofedoren (approved):
Thanks, @adamruzicka !

lib/foreman_puppet/register.rb
```diff
@@ -1,5 +1,5 @@
 Foreman::Plugin.register :foreman_puppet do
-  requires_foreman '>= 3.13'
+  requires_foreman '>= 3.19'
```
nadjaheitmann, line 2: Thanks for fixing! Let's keep in mind to do a major version bump during the next release.

PR #447, opened 2026-06-29
Check tests with theforeman/foreman#11055

adamruzicka:


test/controllers/foreman_puppet/api/v2/host_classes_controller_test.rb
```diff
@@ -20,14 +20,16 @@ class HostClassesControllerTest < ActionController::TestCase
 
         test 'should add a puppetclass to a host' do
           puppetclass = FactoryBot.create(:puppetclass)
-          assert_difference('host.puppet.host_classes.count') do
-            post :create, params: { host_id: host.to_param, puppetclass_id: puppetclass.id }
+          h = host
+          assert_difference(-> { h.puppet.reload.host_classes.count }) do
+            post :create, params: { host_id: host.to_param, puppetclass_id: puppetclass.to_param }
```
adamruzicka:
```
> puppetclass.id
137

> pupetclass.to_param
137-class5
```

hm, I would expect "raw" id's to still work

PR #448, opened 2026-07-17
Fix JS linting error 'no-magic-numbers'

nadjaheitmann:
The Foreman linting requires constants to be declared instead of directly assigning numbers to variables. Otherwise, it will through the 'no-magic-numbers' error.

Corresponding Foreman core PR:

theforeman/foreman#11008

Co-Authored-By: OpenAI Codex


MariaAga (approved):
lgtm! thanks

PR #449, opened 2026-07-17
Fixes #39540 - Host actions are disabled after bulk action

MariaAga:
depends on https://github.com/theforeman/foreman/pull/11110
to test: 
user has permissions to edit hosts
go to new host index
click on host kebab - edit is enabled
run X action
edit is still enabled. 

Disclaimer: I only tested this on foreman core - change location bulk action. since all code looks the same I assume it will still work, but testing is encouraged.

### DIFF
--- a/webpack/src/Extends/Hosts/BulkActions/BulkChangeProxyCommon/index.js
+++ b/webpack/src/Extends/Hosts/BulkActions/BulkChangeProxyCommon/index.js
@@ -15,17 +15,11 @@ import {
 } from '@patternfly/react-core';
 import { addToast } from 'foremanReact/components/ToastsList/slice';
 import { sprintf, translate as __ } from 'foremanReact/common/I18n';
-import { foremanUrl } from 'foremanReact/common/helpers';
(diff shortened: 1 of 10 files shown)


nadjaheitmann:
While I am currently looking at the PR... Do you mind fixing the typo in the second commit message.

nadjaheitmann (approved):
Thanks @MariaAga . Tested and it fixes the mentioned issue. Ready to merge when CI is green.

nadjaheitmann:
As this requires https://github.com/theforeman/foreman/pull/11110, should we also bump the required foreman version to 5.0? Meaning that we are adding a breaking change here and bumping a major version on next foreman_puppet version bump.

nadjaheitmann:
> As this requires [theforeman/foreman#11110](https://github.com/theforeman/foreman/pull/11110), should we also bump the required foreman version to 5.0? Meaning that we are adding a breaking change here and bumping a major version on next foreman_puppet version bump.

I'll merge this to get it in and open a follow-up PR with the foreman dependency

Issue #365, opened 2023-08-01
Search autocomplete on puppetclass throws errors


nadjaheitmann:
In UI, you can see 
`Error: Request failed with status code 500`

In the logs, you can see (not sure what is relevant here)
09:16:11 rails.1   | 2023-08-01T09:16:11 [I|app|fe9dade4] Started GET "/hosts/auto_complete_search?search=puppetclass++%3D" for 192.168.121.1 at 2023-08-01 09:16:11 +0000
09:16:11 rails.1   | 2023-08-01T09:16:11 [I|app|fe9dade4] Processing by HostsController#auto_complete_search as JSON
09:16:11 rails.1   | 2023-08-01T09:16:11 [I|app|fe9dade4]   Parameters: {"search"=>"puppetclass  ="}
09:16:11 rails.1   | 2023-08-01T09:16:11 [I|app|fe9dade4] Completed 500 Internal Server Error in 17ms (ActiveRecord: 3.9ms | Allocations: 9776)
09:16:11 rails.1   |  fe9dade4 | 
09:16:11 rails.1   |  fe9dade4 | 
09:16:11 rails.1   | 2023-08-01T09:16:11 [F|app|fe9dade4]   
09:16:11 rails.1   |  fe9dade4 | NoMethodError (undefined method 'klass' for nil:NilClass):
09:16:11 rails.1   |  fe9dade4 |   
09:16:11 rails.1   |  fe9dade4 | app/models/host/managed.rb:40:in 'complete_for'
09:16:11 rails.1   |  fe9dade4 | app/controllers/concerns/foreman/controller/auto_complete_search.rb:12:in 'auto_complete_search'
09:16:11 rails.1   |  fe9dade4 | app/controllers/concerns/foreman/controller/timezone.rb:10:in 'set_timezone'
09:16:11 rails.1   |  fe9dade4 | app/models/concerns/foreman/thread_session.rb:32:in 'clear_thread'
09:16:11 rails.1   |  fe9dade4 | app/controllers/concerns/foreman/controller/topbar_sweeper.rb:12:in 'set_topbar_sweeper_controller'
09:16:11 rails.1   |  fe9dade4 | /home/vagrant/katello/lib/katello/middleware/organization_created_enforcer.rb:18:in 'call'
09:16:11 rails.1   |  fe9dade4 | /home/vagrant/katello/lib/katello/middleware/event_daemon.rb:10:in 'call'
[...]

PR #450, opened 2026-08-17
Fixes #39620 - Fix Puppet search autocomplete

jakduch:
## Summary

Fixes [#39620](https://projects.theforeman.org/issues/39620) and addresses [GitHub issue #365](https://github.com/theforeman/foreman_puppet/issues/365).

Autocomplete for `puppetclass` and `config_group` raised an exception because their scoped-search definitions referenced regular Ruby helper methods rather than Active Record associations. Scoped-search requires an association reflection to determine the related model when completing values.

Add dedicated through associations for scoped-search metadata and named scopes that select Puppet classes and config groups assigned directly, through host groups, or through config groups. A scoped-search autocomplete builder uses these scopes for the two Puppet fields while leaving all other Foreman completions unchanged. Host group Puppet class search now also includes classes assigned through config groups.

The existing `puppetclasses` helper remains unchanged, including its behavior of building a missing Puppet facet when classes are assigned.

## Testing

- Ruby syntax and whitespace checks pass for all changed files
- Regression tests cover direct, inherited host group, and config group completion paths
- Regression tests exclude unassigned values and preserve Foreman core value completion
- Host group search coverage includes Puppet classes assigned through config groups

## AI usage disclosure

Per the [community discussion on AI policy](https://community.theforeman.org/t/could-foreman-benefit-from-an-ai-usage-policy/44638), the issue was investigated and the changes, tests, and PR wording were prepared with the assistance of Codex 5.6 Sol High. The resulting changes were reviewed before submitting. The commit also includes an `Assisted-By` trailer.

### DIFF
--- a/app/models/concerns/foreman_puppet/extensions/host_common.rb
+++ b/app/models/concerns/foreman_puppet/extensions/host_common.rb
@@ -3,6 +3,41 @@ module ForemanPuppet
     module HostCommon
       extend ActiveSupport::Concern
 
+      ASSIGNMENT_RELATIONS = %i[search_puppetclasses search_config_groups].freeze
+
+      included do
+        singleton_class.prepend CompletionMethods
+
+        has_many :search_puppetclasses, through: :puppet, source: :puppetclasses, class_name: '::ForemanPuppet::Puppetclass'
+        has_many :search_config_groups, through: :puppet, source: :config_groups, class_name: '::ForemanPuppet::ConfigGroup'
+      end
+
+      module CompletionMethods
+        def complete_for(query, options = {})
+          return super if query.nil?
+
+          completer = AssignmentCompleter.new(scoped_search_definition, query, options)
+          return super unless completer.assignment_field?
+
+          completer.build_autocomplete_options
+        end
+      end
+
+      class AssignmentCompleter < ScopedSearch::AutoCompleteBuilder
+        def assignment_field?
+          return false unless complete_options(last_node).include?(:value)
+
+          token = last_token_is(COMPARISON_OPERATORS) ? tokens[-2] : tokens[-3]
+          ASSIGNMENT_RELATIONS.include?(definition.field_by_name(token)&.relation)
+        end
+
+        def completer_scope(field)
+          return super unless ASSIGNMENT_RELATIONS.include?(field.relation)
+
+          field.klass.completer_scope(@options).reorder(Arel.sql(field.quoted_field))
+        end
+      end
+
       def all_puppetclasses(env = environment)
         return ForemanPuppet::Puppetclass.none unless puppet
         puppet.all_puppetclasses(env)
--- a/app/models/concerns/foreman_puppet/extensions/host.rb
+++ b/app/models/concerns/foreman_puppet/extensions/host.rb
@@ -13,9 +13,9 @@ module ForemanPuppet
         include_in_clone puppet: %i[config_groups host_config_groups host_classes]
 
         scoped_search relation: :environment, on: :name, complete_value: true, rename: :environment
-        scoped_search relation: :puppetclasses, on: :name, complete_value: true, rename: :puppetclass, only_explicit: true, operators: ['= ', '~ '],
+        scoped_search relation: :search_puppetclasses, on: :name, complete_value: true, rename: :puppetclass, only_explicit: true, operators: ['= ', '~ '],
           ext_method: :search_by_puppetclass
-        scoped_search relation: :config_groups, on: :name, complete_value: true, rename: :config_group, only_explicit: true, operators: ['= ', '~ '],
+        scoped_search relation: :search_config_groups, on: :name, complete_value: true, rename: :config_group, only_explicit: true, operators: ['= ', '~ '],
           ext_method: :search_by_config_group
       end
 
--- a/app/models/concerns/foreman_puppet/extensions/hostgroup.rb
+++ b/app/models/concerns/foreman_puppet/extensions/hostgroup.rb
@@ -16,13 +16,13 @@ module ForemanPuppet
 
         # will need through relation to work properly
         scoped_search relation: :environment, on: :name, complete_value: true, rename: :environment, only_explicit: true
-        scoped_search relation: :puppetclasses, on: :name,
+        scoped_search relation: :search_puppetclasses, on: :name,
           complete_value: true,
           rename: :puppetclass,
           only_explicit: true,
           operators: ['= ', '~ '],
           ext_method: :search_by_puppetclass
-        scoped_search relation: :config_groups, on: :name,
+        scoped_search relation: :search_config_groups, on: :name,
           complete_value: true,
           rename: :config_group,
           only_explicit: true,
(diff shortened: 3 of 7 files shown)


jakduch:
The Redmine check reports a project mismatch: [#39620](https://projects.theforeman.org/issues/39620) was filed under Foreman, while the affected scoped-search definitions and this fix belong to foreman_puppet. The ticket needs to be moved to the Puppet project by someone with Redmine permissions for the check to pass. [GitHub issue #365](https://github.com/theforeman/foreman_puppet/issues/365) tracks the same bug.

adamruzicka:
Moved the issue in redmine, it should sort itself out on next push

app/models/concerns/foreman_puppet/extensions/host_common.rb
```diff
@@ -3,6 +3,11 @@ module Extensions
     module HostCommon
       extend ActiveSupport::Concern
 
+      included do
+        has_many :search_puppetclasses, through: :puppet, source: :puppetclasses
+        has_many :search_config_groups, through: :puppet, source: :config_groups
```
adamruzicka: Wouldn't this cause the autocomplete to only show a subset of the puppetclasses?

If I understand it correctly, this would offer puppetclasses assigned to the hosts' puppet facets directly, without considering puppetclasses that could be assigned to hosts through hostgroups and config groups.

jakduch: You're right, the new associations only cover classes assigned directly to the Puppet facet. I will rework this and add tests for classes inherited through host groups and config groups.

jakduch: Updated

adamruzicka: I can't say I'm exactly happy about how it turned out, I had to read it a couple of times until it clicked.

Afaik scoped search should be able to leverage named scopes, as long as they return activerecord relations. As a counter proposal, what if we introduced named scopes for puppetclasses and config groups, which would return only the classes/groups which are assigned to any host by any of the possible ways. Then, the autocomplete would offer values from this scope?

It might need some tweaking, but it might end up being cleaner, or at least "less custom". What would you think about this?

jakduch: I stood in a circle, it made sense when I was extending the code but it could be much easier..

app/models/concerns/foreman_puppet/extensions/host_common.rb
```diff
@@ -3,6 +3,49 @@ module Extensions
...
+        def assignment_field?
+          return false unless complete_options(last_node).include?(:value)
+
+          token = last_token_is(COMPARISON_OPERATORS) ? tokens[-2] : tokens[-3]
+          case definition.field_by_name(token)&.relation
+          when :search_puppetclasses, :search_config_groups
+            true
+          else
+            false
+          end
```
adamruzicka: A case is probably a slight overkill both here and in completer_scope below
```suggestion
          [:search_puppetclasses, :search_config_groups].includes?(definition.field_by_name(token)&.relation)
```

app/models/concerns/foreman_puppet/extensions/host_common.rb
adamruzicka: Eh, sorry, I might have misled you slightly with my last comment. This is what I really had in mind: [0001-autocomplete.patch](https://github.com/user-attachments/files/31217635/0001-autocomplete.patch)

app/models/concerns/foreman_puppet/extensions/host_common.rb
jakduch: Updated, both checks now use the shared relation list.

app/models/concerns/foreman_puppet/extensions/host_common.rb
jakduch: Thanks, this is much clearer. I adopted the explicit class names and model completer scopes for both models. I kept the small builder override because scoped_search applies the direct relation filter afterwards; without it, hostgroup- and config-group-only values disappear again. The scopes also merge with super to preserve authorization.

webpack/src/Extends/Hosts/BulkActions/BulkChangeProxyCommon/index.js
```diff
@@ -83,6 +83,7 @@ const BulkChangeProxyCommon = ({
   const toggle = toggleRef => (
     <MenuToggle
       ref={toggleRef}
+      ouiaId="bulk-change-proxy-common-toggle"
```
nadjaheitmann: Personally, I would prefer this one in a separate commit.

jakduch: Done, I split it into a separate commit.

adamruzicka (approved):
I'm fine with this, as long as CI is green

app/models/concerns/foreman_puppet/extensions/host_common.rb
```diff
@@ -3,6 +3,41 @@ module Extensions
...
+          token = last_token_is(COMPARISON_OPERATORS) ? tokens[-2] : tokens[-3]
+          ASSIGNMENT_RELATIONS.include?(definition.field_by_name(token)&.relation)
+        end
+
+        def completer_scope(field)
+          return super unless ASSIGNMENT_RELATIONS.include?(field.relation)
+
+          field.klass.completer_scope(@options).reorder(Arel.sql(field.quoted_field))
+        end
+      end
```
nadjaheitmann: What exactly does this part of the code do?

webpack/src/Extends/Hosts/BulkActions/BulkChangeProxyCommon/index.js
nadjaheitmann: Thanks for addressing my concern, now we merged the fix already in a previous PR, though.

app/models/concerns/foreman_puppet/extensions/host_common.rb
jakduch: This makes autocomplete match the actual Puppet search. Without it, the search handles classes and config groups inherited through host groups, but autocomplete only suggests values assigned directly to a host. This override adds the inherited values for these two fields, while all other fields continue using the default autocomplete. The `reorder` keeps the suggestions sorted by name.

nadjaheitmann: I see, thanks for clarification.

nadjaheitmann:
Ok, tested and seems to work fine for now. What I realized is the following:

- host group search returns the same results as the hosts (returns the puppet classes that are used by hosts)
- the scoped search does not honor taxonomies (was probably broken before?)

I assume that this is still a great improvement compared to the status quo and we can fix this in a follow-up. 

With this thank you @jakduch and merging. Do you mind opening issues for the follow-ups?

jakduch:
Sure, opened [#39683](https://projects.theforeman.org/issues/39683) and [#39684](https://projects.theforeman.org/issues/39684) as follow-ups. Thanks for testing and merging!

PR #454, opened 2026-08-21
Fixes #39684 - Scope Puppet autocomplete by taxonomy

jakduch:
## Summary

Fixes [#39684](https://projects.theforeman.org/issues/39684) as a follow-up to [foreman_puppet #450](https://github.com/theforeman/foreman_puppet/pull/450).

The custom Puppet assignment completer queried classes and config groups globally, bypassing the current organization and location scope of the host or host group being searched. Restrict its assignment subqueries to host and host group Puppet facets visible in the active taxonomies. Direct assignments, inherited host group assignments, and assignments through config groups remain available.

## Testing

- Ruby syntax checks pass for all changed files
- `git diff --check` passes
- Regression coverage verifies organization scoping for host completions and location scoping for host group completions
- Coverage includes both Puppet class and config group values
- Full Rails tests are left to CI

## AI usage disclosure

Per the [community discussion on AI policy](https://community.theforeman.org/t/could-foreman-benefit-from-an-ai-usage-policy/44638), the issue was investigated and the changes, tests, and PR wording were prepared with the assistance of Codex 5.6 Sol High. The resulting changes were reviewed before submitting. The commit also includes an `Assisted-By` trailer.

### DIFF
--- a/app/models/foreman_puppet/config_group.rb
+++ b/app/models/foreman_puppet/config_group.rb
@@ -16,7 +16,9 @@ module ForemanPuppet
 
     validates :name, presence: true, uniqueness: true
 
-    scope :assigned_to_hosts, -> { where(id: ForemanPuppet::HostConfigGroup.select(:config_group_id)) }
+    scope :assigned_to_hosts, lambda {
+      where(id: ForemanPuppet::HostConfigGroup.assigned_to_taxonomy.select(:config_group_id))
+    }
 
     scoped_search on: :name, complete_value: true
     scoped_search relation: :puppetclasses, on: :name, complete_value: true, rename: :puppetclass, only_explicit: true, operators: ['= ', '~ ']
--- a/app/models/foreman_puppet/host_config_group.rb
+++ b/app/models/foreman_puppet/host_config_group.rb
@@ -8,6 +8,18 @@ module ForemanPuppet
 
     validates :host_id, uniqueness: { scope: %i[config_group_id host_type] }
 
+    scope :assigned_to_taxonomy, lambda {
+      host_facets = ForemanPuppet::HostPuppetFacet
+                    .where(host_id: ::Host::Managed.reorder(nil).select(:id))
+                    .select(:id)
+      hostgroup_facets = ForemanPuppet::HostgroupPuppetFacet
+                         .where(hostgroup_id: ::Hostgroup.unscoped.with_taxonomy_scope.reorder(nil).select(:id))
+                         .select(:id)
+
+      where(host_type: ForemanPuppet::HostPuppetFacet.polymorphic_name, host_id: host_facets)
+        .or(where(host_type: ForemanPuppet::HostgroupPuppetFacet.polymorphic_name, host_id: hostgroup_facets))
+    }
+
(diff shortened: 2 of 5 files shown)


nadjaheitmann:
@adamruzicka Can you have a look at this, please? I think we should fix this as revealing information about hosts in another taxonomy could be treated as a potential security issue.

app/models/foreman_puppet/config_group.rb
```diff
@@ -16,7 +16,9 @@ class ConfigGroup < ApplicationRecord
 
     validates :name, presence: true, uniqueness: true
 
-    scope :assigned_to_hosts, -> { where(id: ForemanPuppet::HostConfigGroup.select(:config_group_id)) }
```
nadjaheitmann: I am not sure what is the best way to implement this. Usually, we implement something on the controller side that finds the correct organization and use that information to return the correct data. This is also something that might need attention on the API side, but probably after merging #455

app/models/foreman_puppet/host_config_group.rb
```diff
@@ -8,6 +8,18 @@ class HostConfigGroup < ApplicationRecord
 
     validates :host_id, uniqueness: { scope: %i[config_group_id host_type] }
 
+    scope :assigned_to_taxonomy, lambda {
+      host_facets = ForemanPuppet::HostPuppetFacet
+                    .where(host_id: ::Host::Managed.reorder(nil).select(:id))
+                    .select(:id)
+      hostgroup_facets = ForemanPuppet::HostgroupPuppetFacet
+                         .where(hostgroup_id: ::Hostgroup.unscoped.with_taxonomy_scope.reorder(nil).select(:id))
```
nadjaheitmann: Why do you add 'unscoped' here?

nadjaheitmann:
@jakduch Thanks for submitting the patch! Can you please have a look at the failing tests in the meantime.

app/models/foreman_puppet/host_config_group.rb
jakduch: I followed the same pattern as `Hostext::Search`, but `Hostgroup` already applies the taxonomy scope by default here. I removed `unscoped.with_taxonomy_scope` and kept only `reorder(nil)`. Thanks!

nadjaheitmann:
@adamruzicka Do you mind having a look at this. I am not so familiar with the internals of this but I would really like to move this forward. I don't want to risk having a security issue with the newest foreman_puppet version.

nadjaheitmann:
@adamruzicka Can you please take a look? If you don't have time, I tend to merge this soonish. I have tested the patch locally and it seems to work just fine.

adamruzicka (approved):
> I have tested the patch locally and it seems to work just fine.

I must admit I haven't tested it, but code-wise it looks good and if testing is covered, I wouldn't worry about it.

nadjaheitmann:
> > I have tested the patch locally and it seems to work just fine.
> 
> I must admit I haven't tested it, but code-wise it looks good and if testing is covered, I wouldn't worry about it.

Perfect, thanks, then I'll go ahead and merge it! Thanks @jakduch !

PR #455, opened 2026-08-24
Remove various deprecated API endpoints and parameters

nadjaheitmann:
1) Remove the re-routing to core for puppetclasses, environement and
   config_group endpoints
2) Remove deprecated puppet parameters environement_id, puppetclass_id
   and config_group_id - the parameter group 'puppet_attributes' should
   be used instead


nadjaheitmann:
We should make sure that this does not break the hammer endpoint and FAM.

